PR #8, opened 2017-12-26
Made Dash Background's opacity customizable

nfsprodriver:
Needs to be merged with system-settings' PR

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,8 +13,7 @@ pipeline {
         node(label: 'xenial-arm64') {
           unstash 'source'
           sh '''export architecture="armhf"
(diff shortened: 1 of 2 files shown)


mariogrip:
Thanks for the PR :) but could you target this for the "xenial" branch? since that's where the next ota will be on

mariogrip:
Why did you revert the Jenkinsfile?

nfsprodriver:
Don't know. It appeared when changing the base branch. Jenkins & Co. is your part ;)

mariogrip:
Humm, seems to not touch the jenkins file, so let's merge it and see how it goes

PR #11, opened 2018-02-16
Files and Folders are shown on git status

ismaelbonato:
Remove Files and Folders from git status when I Run unity 8 on my desktop without installing and using Kdevelop4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,14 @@
 /builddir
 # QtCreator project files
 /*.user
+*.qmlc
+*.jsc
 # Eclipse project files
 /.project
 /.cproject
 /.settings
 /.pydevproject
+# KDevelop project Files
+/.kdev4/
+/build/
+/*.kdev4


mardy (approved):
Looks good, thanks!

NeoTheThird:
(not that i don't trust your judgment @mardy :D, i'd just like @mariogrip's two cents on if this is targeting the correct branch and how we want to go about keeping the branches in sync)

@ismaelbonato Awesome, thank you!

mardy:
@NeoTheThird Indeed, whenever I review any submissions I'm paying far more attention to the changes, than to the target branch. So you are right in not hurrying up to merge things :-)

ismaelbonato:
Which branch would be better to do Pull request? This branch is the only I can work properly on my rig for now. But with this initials changes I think it can be applied on the main branch if you want.

mariogrip:
@ismaelbonato Using xenial_-_mir26 or bionic is fine, we will try to keep these two in sync

PR #28, opened 2018-02-25
[In progress] Try to fix 'operation not permitted'

UniversalSuperBox:
When running unity8-dash on the Nexus 5X, I get the error:
Operation not permitted (src/thread.cpp:152)

According to lp#1228201, this works around the problem.


UniversalSuperBox:
Appears that the crash is actually in libunity-scopes

PR #33, opened 2018-03-09
[Stage/Spread] Detect mouse or touch events in newer qt versions

### LINKED ISSUE #487: [xenial] White frame in window spread
### Steps to reproduce

Open the multitasking application fan

### Actual behavior

The highlighted application has a white frame, like when one application is highlighted by hitting tab on the keyboard. Should be de-activated while in touch-screen mode.

mariogrip:
With newer versions of qt a touch event will no longer set "pressed" to
true, this is what the spread has been relying on to detect mouse or
touch events. This adds detection if it's a mouse or touch event in
never versions of qt.

This fixes: https://github.com/ubports/ubuntu-touch/issues/487 and
https://github.com/ubports/ubuntu-touch/issues/430

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -1726,11 +1726,12 @@ FocusScope {
 
         property int scrollAreaWidth: width / 3
         property bool progressiveScrollingEnabled: false
+        property bool isTouch: false
 
         onMouseXChanged: {
             mouse.accepted = false
 
-            if (hoverMouseArea.pressed) {
+            if (hoverMouseArea.pressed || isTouch) {
                 return;
             }
 
@@ -1766,7 +1767,17 @@ FocusScope {
             }
         }
 
-        onPressed: mouse.accepted = false
+        onPressed: {
+          mouse.accepted = false;
+          if (mouse.source === Qt.MouseEventSynthesizedByQt)
+              isTouch = true;
+          else
+              isTouch = false;
+        }
+
+        onExited: {
+          isTouch = false;
+        }
     }
 
     FloatingFlickable {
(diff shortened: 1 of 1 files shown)


mariogrip:
Also using mouse.source in onMouseXChanged or positionChanged will just give MouseEventNotSynthesized (aka mouse event) for some odd reason. oddly the first event is correct but later on it changes to MouseEventNotSynthesized...

qml/Stage/Stage.qml
```diff
@@ -1766,7 +1767,17 @@ FocusScope {
             }
         }
 
-        onPressed: mouse.accepted = false
+        onPressed: {
+          mouse.accepted = false;
+          if (mouse.source === Qt.MouseEventSynthesizedByQt)
+              isTouch = true;
+          else
+              isTouch = false;
```
dpniel: Could we get this down to a single line `isTouch = mouse.source === Qt.Mouse...` just to make it a bit cleaner

qml/Stage/Stage.qml
```diff
@@ -1726,11 +1726,12 @@ FocusScope {
 
         property int scrollAreaWidth: width / 3
         property bool progressiveScrollingEnabled: false
+        property bool isTouch: false
```
dpniel: Would `wasTouchPress` be a better name or something like that to indicate the last event was a touch event? I'm fine with it as it is though if it just had little comment to state what it represents

dpniel (changes requested):

qml/Stage/Stage.qml
mariogrip: Yeah sorry about that, had it like this since i had some console.log's in there to debug with.

qml/Stage/Stage.qml
mariogrip: Jep

mariogrip:
Pushed a commit

dpniel (approved):
LGTM

PR #41, opened 2018-06-07
Do not store window size of zero-width window

doniks:
https://github.com/ubports/ubuntu-touch/issues/460

### DIFF
--- a/qml/Stage/WindowResizeArea.qml
+++ b/qml/Stage/WindowResizeArea.qml
@@ -101,7 +101,9 @@ MouseArea {
             state = WindowStateStorage.WindowStateNormal;
         }
         windowStateStorage.saveState(root.windowId, state & ~WindowStateStorage.WindowStateMinimized); // clear the minimized bit when saving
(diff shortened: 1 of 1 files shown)


qml/Stage/WindowResizeArea.qml
```diff
@@ -101,7 +101,10 @@ MouseArea {
             state = WindowStateStorage.WindowStateNormal;
         }
         windowStateStorage.saveState(root.windowId, state & ~WindowStateStorage.WindowStateMinimized); // clear the minimized bit when saving
-        windowStateStorage.saveGeometry(root.windowId, Qt.rect(priv.normalX, priv.normalY, priv.normalWidth, priv.normalHeight));
+        if ( priv.normalWidth !== 0 )
```
dpniel: should `normalHeight` also be checked here? as either way we don't them to be stored as 0.

Also this could be shortened to `if (priv.normalWidth) {` as 0 equates to false.

dpniel (changes requested):

doniks:
messed up my branch, will pr again

PR #60, opened 2018-08-11
Add support for red filter overlay

mariogrip:


nfsprodriver:
Maybe someone can help to optimize that time thing and make the Shell continously compare the time in the Shell part...

nfsprodriver:
Sorry for confusing, I don't have got a real idea how rebasing works yet. It's a indivdual WIP :$

NeoTheThird:
Is this ready now? Do you want us to go ahead with testing?

nfsprodriver:
The time thing isn't still working yet. I don't know, how to get a time trigger (signal?) working...

NeoTheThird:
@mariogrip any ideas?

nfsprodriver:
Does anyone know how to log the Shell.qml outputs? .cache/upstart/unity8.log doesn't offer me information.

dobey:
A few comments on this. Firstly, please also include an "always on" option for the overlay. Requiring the filter to be off for at least 1 minute in order to get close to always on, is super annoying in Android, and not something we should replicate.

Secondly, would it not be better to move the comparison logic into C++ so you can use the more advanced QTime API to more simply compare the time values directly? There's a lot of if branches in that JS, which would be good to avoid for performance reasons, if possible.

Thirdly, your Timer is not running because nothing ever starts it. You probably need to add a Component.onCompleted: overlayTimer.start() or similar inside the overlay component.

nfsprodriver:
@dobey Thanks for your hints! I'm planning to replace the diff() function soon.

nfsprodriver:
Does that work for anyone yet?

ma3yta (approved):

UniversalSuperBox:
This filter overlay is far too slow to be used on real devices. We'll need to find another way to filter color that is faster... maybe in the display server?

PR #62, opened 2018-08-15
Don't just skip if we are not connected and dont skip twice if not supported

mariogrip:
This result in a "white empty page" since we skip past finished page


NeoTheThird (approved):
Afaik, this seems OK

PR #66, opened 2018-08-24
Disable firmware update page pending more testing

UniversalSuperBox:
This PR removes the firmware update page from navigation in the Wizard. We'll add it back after OTA-4.

Test with ubports/system-settings#77


mariogrip (approved):
LGTM

PR #73, opened 2018-10-01
Don't fail silently when qdbusxml2cpp is missing

z3ntu:
For some reason `execute_process` in CMake doesn't care if a command doesn't exist and doesn't even say it failed - so check immediately if qdbusxml2cpp is available.


z3ntu:
~~Probably breaks for Debian... Will adjust tomorrow~~

z3ntu:
Actually it doesn't because both code paths set the variable.

mariogrip (approved):
LGTM

PR #75, opened 2018-10-06
Purge dash and scopes dependency.

dobey:
Remove the dash app and scopes related code.


Flohack74:
I cannot review all single files, what you suggest ;)

dobey:
I don't know. It's 99.999% deleting code anyway. If you run unity8 on 18.04 anywhere, I suppose you can install and test. I've installed this and the url-dispatcher changes in my VM, and purged libunity-scopes1.0 to test that everything else still works as expected for example. I may have missed a few more things we can probably remove too. I guess if GitHub had a "hide deleted files" option in the diff view, it'd be easier to review. This is I guess mostly just a request for core devs to sign off on getting this done on bionic now, so it can be easier for people to get unity8 packages made for their distros.

Flohack74:
Sure I trust you that you wont break it anyway ;) I approve, but I would ask someone who has 18.04 running to make a live test.

Flohack74 (approved):

PR #76, opened 2018-10-08
Fix more wizard papercuts

UniversalSuperBox:
- It's not possible to skip the app update page while it's checking for
updates, so make it say "Please wait..." while checking
- The Changelog page now indicates that the next page is loading.
This prevents people from clicking "Next" multiple times and skipping
the app update page.


dobey (approved):
Looks OK to me.

PR #77, opened 2018-10-10
Disable apport

mariogrip:


dobey:
What's the goal here? This change seems like an odd/inappropriate way to achieve something more specific. I thought we already got rid of whoopsie as well, or did I miss something there?

mariogrip:
@dobey this requires this though https://github.com/ubports/ubuntu-touch-meta/pull/30

It is to remove the whoopsie function

dobey:
OK. I think this specific change is the wrong way to remove apport though. Won't renaming the QML file here just break whatever is importing/loading it?

Also you'lll want https://github.com/ubports/system-settings/pull/88

mariogrip:
@dobey It allows you to disable wizard pages by adding .disabled at the end of it: https://github.com/ubports/unity8/blob/xenial/plugins/Wizard/PageList.cpp#L45

dobey:
@mariogrip Right, and that's perhaps useful for testing periods. But I think if we want to remove stuff we should be ripping it out, not just renaming things to disabled or commenting lines in code. It's in git history, so no need to keep things around we won't need, and have confusing commented blocks of code or disabled files with no direct info on why they are that way.

I can help to remove it more cleanly if you need me to.

mariogrip:
@dobey sure, i guess we wont get used for it anytime soon anyway, and as you say we got git history so yeah, just rm it all.

UniversalSuperBox (approved):

PR #79, opened 2018-10-14
Advance welcome wizard when pressing enter in text fields

nikalon:
Fixes #846

### DIFF
--- a/qml/Wizard/Pages/60-account.qml
+++ b/qml/Wizard/Pages/60-account.qml
@@ -26,9 +26,13 @@ LocalComponents.Page {
     forwardButtonSourceComponent: forwardButton
     onlyOnInstall: true
 
-    QtObject {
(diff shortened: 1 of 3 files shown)


UniversalSuperBox (approved):
This tests good to me. I can do all the following successfully:

* On the name page:
  * Press Enter after typing a name to advance
  * Press Next after typing a name to advance
* On the Password page:
  * Press Enter to advance from "password" to "repeat password"
  * Tap on both fields independently to focus them
  * Press Enter to advance off the page once both fields are valid and match
  * Press Next to advance off the page once both fields are valid and match

And, as expected, my name and password are set in both cases. :)

This is a rockstar first contribution! Be sure to sign up for [Hacktoberfest](https://hacktoberfest.digitalocean.com) if you haven't already so this counts toward your swag!

PR #83, opened 2018-11-03
Update com.canonical.Unity8.gschema.xml

kugiigi:
Changed "edge-barrier-sensitivity" from "y" to "u"
UT Tweak Tool has an error setting this schema key because it is set to "y" (guchar).
Other keys that involves a numerical range in Unity8 are set to "u" so maybe this is the correct one for "edge-barrier-sensitivity" as well.

### DIFF
--- a/data/com.canonical.Unity8.gschema.xml
+++ b/data/com.canonical.Unity8.gschema.xml
@@ -11,7 +11,7 @@
       <summary>The usage mode.</summary>
       <description>The usage mode chosen will affect the Window Management behaviour.</description>
     </key>
-    <key type="y" name="edge-barrier-sensitivity">
+    <key type="u" name="edge-barrier-sensitivity">
       <default>35</default>
       <range min="1" max="100"/>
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
Can this change be applied without any migration? Does unity8 continue to work correctly?

kugiigi:
> Can this change be applied without any migration? Does unity8 continue to work correctly?

I'm not exactly sure but when I changed it on my device, it was working fine.

UniversalSuperBox (approved):
Seems like this is a compatible type change, so let's do it!

UniversalSuperBox:
For QA, the main thing is to ensure that edge swipe behavior is as expected. If you had a custom value set before, check if it's still set properly.

jezek:
FP2 (2019-W18)
Edge swipe behavior is as expected, no change.

ziggutas:
BQ E4.5 rc (2019-W18)
Edge swipe same as before

ziggutas:
Nexus 7 (flo) rc (2019-W18)
Same as before

PR #91, opened 2019-01-07
Fixup u8 tests and remove oxide to make arm64 happy

mariogrip:


mariogrip:
*on hold* 
lets wait until we know oxide wont get removed (from edge's rootfs), if so we want to add oxide to armhf meta package.

plugins/Wizard/System.cpp
```diff
@@ -119,7 +119,11 @@ void System::setWizardEnabled(bool enabled)
         if (QDir(wizardEnabledPath()).exists()) {
             QDir(wizardEnabledPath()).removeRecursively();
         }
-        QFile::copy(currentFrameworkPath(), wizardEnabledPath());
+        if (!QFile::copy(currentFrameworkPath(), wizardEnabledPath())) {
+          // Make en empty file if framework does not exist
+          QFile f(wizardEnabledPath());
+          f.open(QFile::WriteOnly);
```
dobey: Please match indentation alignment with the rest of the file for consistency (should be 4 spaces inside if block as seen above this change). :)

dobey (changes requested):

UniversalSuperBox (approved):
I've ensured that this won't remove oxide from the image and updated the indentation. Waiting for @dobey to re-check

dobey (approved):

PR #92, opened 2019-01-24
Automatically advance setup wizard on WiFi connection

### LINKED ISSUE #845: Welcome Wizard: Should automatically advance after network connection
When selecting a wifi network in the welcome wizard it would be nice if it automatically advanced to the next page after a network connection is made.

Arc676:
Employs the fix suggested by @UniversalSuperBox to automatically advance the setup wizard once a connection to a WiFi network is established.

Tested by editing the file locally and re-running the setup wizard on Nexus 5 (hammerhead) running 16.04 (2019-01-11), `devel` channel.

Closes ubports/ubuntu-touch#845.

### DIFF
--- a/qml/Wizard/Pages/30-wifi.qml
+++ b/qml/Wizard/Pages/30-wifi.qml
@@ -34,6 +34,12 @@ LocalComponents.Page {
     readonly property bool connected: Connectivity.online
     skip: connected
 
+    onConnectedChanged: {
+        if (connected) {
+            pageStack.next()
+        }
+    }
+
     function getExtendedProperty(object, propertyName, defaultValue) {
         if (object && object.hasOwnProperty(propertyName)) {
             return object[propertyName];
--- a/qml/Wizard/Pages/30-wifi.qml
+++ b/qml/Wizard/Pages/30-wifi.qml
@@ -35,7 +35,7 @@ LocalComponents.Page {
     skip: connected
 
     onConnectedChanged: {
-        if (connected) {
+        if (connected && wifiPage.visible) {
             pageStack.next()
         }
     }
--- a/qml/Wizard/Pages/30-wifi.qml
+++ b/qml/Wizard/Pages/30-wifi.qml
@@ -32,7 +32,7 @@ LocalComponents.Page {
     forwardButtonSourceComponent: forwardButton
 
     readonly property bool connected: Connectivity.online
-    skip: connected && !Connectivity.limitedBandwidth && wifiPage.visible
+    skip: connected && !Connectivity.limitedBandwidth
 
     onConnectedChanged: {
         if (connected && !Connectivity.limitedBandwidth && wifiPage.visible) {
--- a/qml/Wizard/Pages/30-wifi.qml
+++ b/qml/Wizard/Pages/30-wifi.qml
@@ -32,7 +32,7 @@ LocalComponents.Page {
     forwardButtonSourceComponent: forwardButton
 
     readonly property bool connected: Connectivity.online
-    skip: connected
+    skip: connected && !Connectivity.limitedBandwidth && wifiPage.visible
 
     onConnectedChanged: {
         if (connected && !Connectivity.limitedBandwidth && wifiPage.visible) {
--- a/qml/Wizard/Pages/30-wifi.qml
+++ b/qml/Wizard/Pages/30-wifi.qml
@@ -35,7 +35,7 @@ LocalComponents.Page {
     skip: connected
 
     onConnectedChanged: {
-        if (connected && wifiPage.visible) {
+        if (connected && !Connectivity.limitedBandwidth && wifiPage.visible) {
             pageStack.next()
         }
     }


Arc676:
Looks like one of the test cases (`testWizardSystem`) isn't expecting this behavior, unless that test case was already failing before.

UniversalSuperBox (changes requested):
You should also check if this is the currently active page, else re-connections will cause the next pages to be skipped.

UniversalSuperBox:
Another tip, you can check if the page is visible or not using the `visible` property of any item on the page.

Arc676:
Does the `visible` property match whether the page is active? Or can a page be visible behind another view?

Sorry for the inactivity; currently on my other partition. I'll probably switch back to Ubuntu to test this soon.

UniversalSuperBox:
The [`visible` property](https://doc.qt.io/qt-5/qml-qtquick-item.html#visible-prop) is set for each page, and it propagates to every item on the page when set. Visible is set to true when the page is focused, false when it is left.

Arc676:
Added additional condition that `wifiPage` must be visible.

Performed the following tests and observed the following behavior by turning the personal hotspot on my other phone on and off at various times:

- If the device is connected to a network before reaching the WiFi screen, that screen is skipped
- If the device becomes connected to a known network without user interaction, the language screen is not skipped
- If the user manually connects to a network, there is no need to press the "next" button; the device setup progresses automatically
- If the device becomes connected while the WiFi screen is active without user interaction, the screen is automatically skipped

It's worth noting that due to the network refresh rate, networks that suddenly become available like in this test will not immediately appear. I had to wait a few seconds for the device to connect to my hotspot after I turned it on while on the WiFi screen. But this seems like an unlikely scenario for most end users.

UniversalSuperBox:
While testing in a device with a SIM inserted, I realized another problem -- any data connection triggers `Connectivity.online`...

Something must know if Wi-Fi is connected or not, since updates only check over Wi-Fi. Let's investigate more...

UniversalSuperBox:
On further investigation (read: poking Rodney), `Connectivity.limitedBandwidth` should be false for Wi-Fi connections.

Arc676:
Gotcha. Added the extra condition for WiFi connections. I can confirm that connecting to a network still skips the network selection screen (specifically the fourth scenario in my last comment). Unfortunately I don't have a SIM card in my device so I can't check mobile data connections.

UniversalSuperBox:
Oh, I didn't think of putting it there! That fixes a new bug that I would have introduced were I making the changes.

That condition also needs to be added to `skip: `

Arc676:
Out of curiosity (and also for a better understanding of the code), where were you thinking of making the changes and what was the bug?

UniversalSuperBox:
This page gets automatically skipped whenever you are connected to any network because the `skip: ` property is just set to `Connectivity.online`. I mistakenly attributed that to your changes.

You don't have to change it if you don't want to, but I'd appreciate it.

Arc676:
I see. I changed `skip` to be the exact same expression as the skip condition in c6941b0. Still tested and observed same behavior (although again with no mobile network test).

UniversalSuperBox:
Alright, now I see what's going on here better. `skip` is used to decide whether the page should be skipped *before* it is loaded. If `skip` is true, the page will never be visible. So, one more change is needed.

Arc676:
I must admit I'm not sure what the necessary change is. I'm assuming for readability, it shouldn't include `wifiPage.visible`, but this didn't affect functionality. Regardless of whether the device was already connected before beginning setup or became connected before reaching the network selection page, the page was always skipped.

What am I missing?

UniversalSuperBox:
`wifiPage.visible` should always be false when it's evaluated (it's deciding whether to show the page or not), so the page is always shown.

Arc676:
That's very weird because the page got skipped on my device when it was connected before reaching the network selection screen. I'll try testing it again tonight.

Arc676:
Tested by turning on the hotspot on my other phone while the language selection screen was present. The device connected automatically to the network and after hitting "skip" on the "no SIM card" screen, it jumped straight to the time zone selection.

How might I be able to debug this?

UniversalSuperBox:
Oh, that's a strange behavior. Pages get set to `visible` before they are made visible, while the Wizard is trying to decide whether to skip them or not.

So yes, you're right, this works normally... however it relies on `visible` being set when it really shouldn't be (the page isn't being shown, we're trying to figure out if we should skip it!). Even though it isn't really necessary, I'd prefer if `&& visible` was removed from the skip property.

Arc676:
Makes sense. Future-proofing the condition it is then, in case a future version doesn't set visible to `true` before checking whether to skip. I've removed the condition.

UniversalSuperBox (approved):
And after all that, it seems like `Connectivity.limitedBandwidth` doesn't actually become true when connected to 4G... oh well, that's no reason to keep this from the codebase any longer. The skip condition works as designed, as does the automatic advance.

Thank you so, so much for working with me on this. I've learned a lot about the Wizard.

UniversalSuperBox:
Would you like to be mentioned by name (as listed on GitHub) in our changelog? The changelog will appear on the post releasing OTA-9, similar to the OTA-8 post: https://ubports.com/blog/ubports-blog-1/post/ubuntu-touch-ota-8-release-207

Arc676:
It's been a pleasure. Thanks to you as well for all the input! And coming up with the original fix :) Hopefully with this as a base I might be able to better understand the codebase and make more significant contributions in the future.

Per the changelog: It'd be an honor to be mentioned. Thanks!

PR #93, opened 2019-02-08
Add a timeout for snapping a screenshot.

### LINKED ISSUE #1015: Feature Request: Disallow the taking of screenshots of lockscreen.
I am using a OPO (stable, OTA-5) and was also experiencing this issue on a Nexus 5 (stable, OTA-4).

In short, I am generating about 2-15 unintentional screenshots per day. What follows may be an opinion, but having to press two buttons that are next to and are ATTACHED to each other will intuitively result in this problem. SailfishOS uses the same button combo, so maybe there is some annoying trademark or patent issue that I am not thinking about that is the reason this particular decision was made. The number of screenshots are small, so this is nothing but an annoyance.

As no one is complaining about this issue, and I have experienced it frequently on two UT devices, maybe I am just using my phone in a really wacky way. It usually happens as I am fumbling with the phone as I take it out of my pocket, when I pull it out upsidedown or when I am carrying the phone in my hand while trying to do other things. I also have a lot of screenshots of times when I have attempted to change the volume.

I use a two-layer case, that has covered buttons which is likely making the issue worse, but it also happens without a case on both devices. I have also **NEVER** unintentionally taken a screenshot on any smartphone running any other OS with any case.

Short of changing the button combination to ANYTHING else (Power + Vol Up?, Power + Vol Down?, Vol Up + Vol Down + Finger on screen?), the thing that would solve the bulk of this problem would be to disallow the taking of a screenshot while the phone is locked. I for one will never in a million years INTEND to take a screenshot of my lockscreen. There may be reasons to, but I can't think of one

dobey:
Wait for half of the pressAndHoldInterval before signalling to take a
screenshot. This is to prevent accidental screenshots from unintended dual
pressing of the volume buttons. Fixes ubports/ubuntu-touch#1015


UniversalSuperBox (changes requested):
This doesn't seem to generate the desired behavior on my Fairphone... I must press and hold the buttons, then press them again to generate a screenshot. It's difficult to do this without triggering the normal volume change action.

qml/Components/PhysicalKeysMapper.qml
```diff
@@ -45,13 +45,21 @@ Item {
     readonly property bool superPressed: d.superPressed
     readonly property bool superTabPressed: d.superTabPressed
 
+    // Add a timeout for screenshots, but let it be half of normal long press
+    readonly property int screenshotTimeout: mouseArea.pressAndHoldInterval / 2
```
mardy, line 49: You don't need the MouseArea, you can use `Qt.styleHints.mousePressAndHoldInterval`

qml/Components/PhysicalKeysMapper.qml
```diff
@@ -89,21 +98,31 @@ Item {
             if (event.isAutoRepeat && !d.ignoreVolumeEvents) root.volumeDownTriggered();
             else if (!event.isAutoRepeat) {
                 if (d.volumeUpKeyPressed) {
-                    if (Powerd.status === Powerd.On) {
+                    if (d.screenshotPressStart > 0
+                        && currentEventTimestamp - d.screenshotPressStart >= root.screenshotTimeout
```
mardy, line 102: You'll need to move this out of the `(!event.isAutoRepeat)` condition somehow, or it will never be hit. See how the PowerDown/PowerOff keys above are handled.

mardy (changes requested):

Flohack74:
@UniversalSuperBox that would be a late candidate for OTA-9 if @dobey has time to look at it again?

dobey:
> @UniversalSuperBox that would be a late candidate for OTA-9 if @dobey has time to look at it again?

I don't think I will have time to resolve this before OTA9 freeze, unfortunately.

NeoTheThird:
Since OTA-11 is supposed to be mostly PRs that have been shelved for a while, maybe we could look at this one again? Or is it no longer needed?

Flohack74:
Sure, lets try...

NeoTheThird:
Removing again, not really a high priority right now, so it should not halt the ota.

UniversalSuperBox:
This PR has become stale and now conflicts with its base branch, so I'll close it.

PR #98, opened 2019-02-21
Remove the blur effect from the Drawer :( 

### LINKED ISSUE #97: Blur on app drawer is too slow for phones
The blur on the app drawer looks awesome on phones, but unfortunately it is too slow during the swipe-in/out animation. Apparently Qt 5.10 fixed this by caching the shader used to make the blur, but we're still on 5.9. Until we upgrade, we'll need to remove the blur.

UniversalSuperBox:
These commits remove the blur effect from the Unity8 Drawer and make its background transparency extremely slight. This gives a small interesting effect while not being overbearing on bright backgrounds with text.

image

I mixed up "Launcher" and "Drawer" in the commit title... Maybe squash this and say "Drawer" rather than "Launcher"

Fixes #97


qml/Launcher/Drawer.qml
```diff
@@ -69,7 +69,8 @@ FocusScope {
 
     Rectangle {
         anchors.fill: parent
-        color: "#BF000000"
+        color: theme.palette.normal.background
+        opacity: 0.99
```
dobey, line 73: "BF" is 0.6875, not 0.99. Seems like opacity being 0.99 might as well be 1.0 at that point. Probably 0.6875 is a bit too transparent though, so maybe something closer to 0.90 would be best? Maybe something between 0.85 and 0.93?

dobey: Should also probably always use "#000000" for the color here, so we don't end up having it be white in some odd case in the future?

UniversalSuperBox: I tried that, but the transparency was still really distracting when the background was white (so, every app in Ubuntu Touch right now).

dobey (approved):

UniversalSuperBox:
There, the background is SuruDark's background color now. Does that work better?

PR #100, opened 2019-02-23
Add backup method for detection of PC/mobile if hostname.chassis is empty

mariogrip:
On some devices hostname.chassis is empty, this adds a backup method of
figuring out if we are on mobile or not by checking if lxc-android-config is installed

Fixes popup of power dialog on single press of btn

### DIFF
--- a/plugins/Unity/Platform/platform.cpp
+++ b/plugins/Unity/Platform/platform.cpp
@@ -19,6 +19,7 @@
 #include <QDBusConnection>
 #include <QSet>
 #include <QString>
+#include <QFileInfo>
 
 Platform::Platform(QObject *parent)
     : QObject(parent), m_isPC(true), m_isMultiSession(true)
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
Which devices do not have hostname.chassis set?

UniversalSuperBox:
Ah, I see. The configs for the FP2 and Nexus 5 in `/etc/ubuntu-touch-session.d` don't have the FORM_FACTOR environment variable set. See https://docs.ubports.com/en/latest/porting/running-ut.html#form-factor for a true fix.

mariogrip:
oh, I was looking for something like that. but i guess thats something canonical added maybe? But I still think this is a valid fix *if its empty* since i can see this happening on other halium devices. It checks for lxc-android-config, and if its there default to "phone" if not default to "desktop".  

The old behavior was check if == "desktop" but they changed it to if != "mobile" and thats why this happens now. So before it defaulted to "phone" now to "desktop"

PR #101, opened 2019-02-23
Add backup method for detection of PC/mobile if hostname.chassis is empty

mariogrip:
If FORM_FACTOR is not set hostname.chassis will be empty, this adds a backup method of figuring out if we are on mobile or not by checking if lxc-android-config is installed, if its installed, default to phone otherwise desktop

This makes it easier for porters, this is how the old behavior was, where it checked if == "desktop" but it's now changed it to if != "mobile", aka before it defaulted to "phone" now to "desktop"

See: https://github.com/ubports/unity8/pull/100

### DIFF
--- a/plugins/Unity/Platform/platform.cpp
+++ b/plugins/Unity/Platform/platform.cpp
@@ -19,6 +19,7 @@
 #include <QDBusConnection>
 #include <QSet>
 #include <QString>
+#include <QFileInfo>
 
 Platform::Platform(QObject *parent)
     : QObject(parent), m_isPC(true), m_isMultiSession(true)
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
I don't agree with adding this right now... it locks Unity8 to Ubuntu (Touch) more than necessary. For a similar shim (if lxc-android-config is installed, select phone), we could add the phone chassis type to https://github.com/ubports/ubuntu-touch-session/blob/xenial/ubuntu-touch-session.d/android.conf. That file is loaded whenever no bind-mounted file is provided.

UniversalSuperBox:
Thinking about this more, might it be better to leave the session file untouched and have Unity8 start incorrectly? That would enforce the use of appropriate config files to configure Ubuntu Touch on each device.

PR #103, opened 2019-02-25
Build-depend directly on android-headers

UniversalSuperBox:
It does what it says on the tin and fixes the build in the cmake configuration


dobey (approved):
Ok.

Issue #110, opened 2019-03-19
Clear searches when the Drawer is closed


UniversalSuperBox:
This is a design change from the current behavior of the App Scope and Unity7.

When the Drawer is closed via any method (launching an app, tapping Escape, swiping it away), the user's search is not cleared. When opening the Drawer again, all of the text in the search box is selected but not cleared.

This can lead to situations such as the following:

1. I type "Gal"
1. I launch the Gallery
1. After using the Gallery, I want to use the Calculator. I open the Drawer
1. "Gal" is still in the search box and the app list is still filtered.

In normal use, I expect my search to be cleared when I leave the Drawer. This expected behavior occurs in Windows, macOS, Android, iOS, Plasma, GNOME... when you leave a launcher containing a search box, it is cleared.

UniversalSuperBox:
Closed by #112

Issue #107, opened 2019-03-18
Opening the Drawer the first time causes OSK to open


UniversalSuperBox:
When opening the Drawer or Launcher with the OSK enabled, the OSK appears immediately upon releasing the gesture. This only occurs the first time that these are opened after Unity8 starts.

This appears to be related to the [focusInput function](https://github.com/ubports/unity8/blob/7f172ec231a50527e970103128dbbde8de2bcec4/qml/Launcher/Drawer.qml#L44) being called at an inappropriate time... Whenever a text field gains focus, the keyboard appears. The art for us is to ensure that the text field is immediately focused when text is entered using a hardware keyboard without making it focused immediately on opening.

UniversalSuperBox:
Closed by #109

PR #112, opened 2019-03-19
Clear the search when the Drawer is closed

UniversalSuperBox:
I ended up fighting with keyboard focus while writing this commit, so I
had to refactor how the Launcher and Drawer handle keeping focus on
the Drawer's search field.

Closes https://github.com/ubports/unity8/issues/110
Fixes some other behavior related to https://github.com/ubports/unity8/issues/107

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -35,12 +35,6 @@ FocusScope {
     property bool draggingHorizontally: false
     property int dragDistance: 0
 
-    onFocusChanged: {
-        if (focus) {
-            searchField.selectAll();
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
I've removed the use of `clear()`, which I was working on adding to the UITK... I'm not sure it's worth it to do API changes to the UITK, rather we should work on other projects improving upstream technologies like QQC2.

dobey (approved):
Looks good. It's a bit upsetting that no tests broke, so I wonder if that's because there are no tests for this, or if they aren't getting run for some reason. Would be nice to get coverage for all this to ensure we don't break it in the future.

PR #114, opened 2019-03-26
Never propagate composed events

UniversalSuperBox:
There's a MouseArea in the Drawer which detects when the user drags horizontally to respond properly. For some reason, when its `propagateComposedEvents` property is toggled in the way before these changes, it fails to let through the first scrolling action after the Drawer is opened. When it's always false, it can still detect horizontal scrolling and lets through the scrolling action. Nothing else seems to be affected.

I have no idea why it didn't work before or why it works now.

Fixes https://github.com/ubports/ubuntu-touch/issues/1080, somehow.

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -190,9 +190,10 @@ FocusScope {
             }
 
             MouseArea {
+                id: horizontalDragDetector
                 parent: listLoader.item ? listLoader.item : null
                 anchors.fill: parent
-                propagateComposedEvents: true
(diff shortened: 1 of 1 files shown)


qml/Launcher/Drawer.qml
```diff
@@ -190,9 +190,10 @@ FocusScope {
             }
 
             MouseArea {
+                id: horizontalDragDetector
```
fredldotme: Do you need the id if it is never referenced?

dobey (approved):
Looks ok to me.

PR #116, opened 2019-03-27
Bug swap

UniversalSuperBox:
It seems like, for the first interaction with Unity8's Drawer in
a session, we can pick any two:
* Horizontal drag (dismiss)
* Vertical drag (scroll)
* Click (Launch)

This is due to what may be a bug where a Flickable won't steal the mouse event from its parented MouseArea... but only the first event after its creation.

Our previous behavior allowed horizontal drag and click interactions.
This change allows vertical drag and click interactions.
Swapping one bug for another.

This is honestly a hack, but logic tells me that the two behaviors which are allowed with these changes are more likely than the dismiss action as the first thing someone does after launching Unity8. I understand if we don't want to merge it until absolutely necessary (after investigating and submitting a bug upstream).

Fixes https://github.com/ubports/ubuntu-touch/issues/1080

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -32,6 +32,7 @@ FocusScope {
 
     signal applicationSelected(string appId)
 
+    property bool firstOpen: true
     property bool draggingHorizontally: false
     property int dragDistance: 0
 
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
Flickables not stealing the first drag off their child has been reported as https://bugreports.qt.io/browse/QTBUG-74842

dobey (approved):

Issue #108, opened 2019-03-18
The Drawer's sorting takes up too much space on mobile


UniversalSuperBox:
The current default sort on the Drawer is "A-Z", which separates apps by the first letter in their name. This causes a couple of problems:

1. There are many categories with only one app, so we're using a lot of vertical space
1. In non-english languages, the first localized character of the app name is used. This causes even more groups to be created (one for localized names, another for non-localized English names). Languages with many letters weep.

Our current idea to solve this problem is to display all apps, sorted alphabetically, in a single group.

Here is the current design (on the Nexus 7):

![Nexus 7 Drawer](https://user-images.githubusercontent.com/21966173/54564579-1d7c7e00-499a-11e9-903f-8055b610dc71.png)

mariogrip:
Fixed by: https://github.com/ubports/unity8/pull/120

PR #120, opened 2019-04-01
Remove the grouping from the Drawer

UniversalSuperBox:
To simplify the Drawer code, I've removed all of the ListViews. This gets rid of the ability to have separate sorting options, however we can always bring this back later if we wish. Now the only available sorting is A-Z with no groups.

If nothing else, the AppDrawerProxyModel is unchanged, so we can always implement this again (possibly cleaner?) later.

Currently I'm working out the keyboard focus, so this PR is in draft status.

Fixes https://github.com/ubports/unity8/issues/108


NeoTheThird:
Love it! But maybe tapping the bfb should only close the drawer and not the launcher as well?

NeoTheThird:
Especially since you even close the Launcher when it's supposed to always be displayed (eg window-mode with always-show-launcher option). That needs to be fixed.

UniversalSuperBox:
Oops. It'll be at least Friday before I'm able to look into that (or the keyboard stuff), thanks for bringing it up though.

UniversalSuperBox:
Well, that should fix that up.

Something to note: When the grid view has keyboard focus, the blue focus indicator gets clipped by the top of the Flickable. I've spent more time than I should fixing it now, but otherwise have keyboard navigation nailed down pretty good.

NeoTheThird:
Seems OK. Good job!

Did you want for the drawer to be the same color as the launcher? I guess the blur thing is still planned to come back at some point, when we figured out the performance, but maybe the color should be a little different? Don't know.

Another thing, what about tests? *dodges the rotten tomatoes*

qml/Launcher/LauncherPanel.qml
```diff
@@ -97,7 +98,7 @@ Rectangle {
                 sourceSize.width: width
                 sourceSize.height: height
                 anchors.centerIn: parent
-                source: "graphics/home.svg"
+                source: root.showCloseIcon ? "graphics/back.svg": "graphics/home.svg"
```
dobey: Could we make this an `Icon` instead of `Image`? We could drop the `sourceSize` settings here, as that's already handled by `Icon`, and use `name: root.showCloseIcon ? "back" : ""` along with `source: root.showCloseIcon ? "" : "graphics/home.svg"` and avoid having to keep a copy of the icon in unity8 itself, I think. And maybe set `color` property on it, so the back button also gets the right color when it's on.

UniversalSuperBox:
```
Icon {
    objectName: "dashItem"
    width: parent.width * .6
    height: width
    anchors.centerIn: parent
    name: root.showCloseIcon ? "back" : ""
    source: root.showCloseIcon ? "" : "graphics/home.svg"
    color: white
    rotation: root.rotation
}
```

Seems to work for the Ubuntu icon, but it doesn't load Back correctly. Something I did wrong here?

dobey: No. I misremembered how the source vs name usage there works. :-/

UniversalSuperBox: So... that idea won't work then?

dobey: I'm not sure if something slightly more complex might work or not, without a couple simple changes to `Icon` source. Maybe just include the full path to the SVG in `/usr/share/icons/` instead of shipping a copy, for now?

dobey: That is, don't use `name` property at all, but just the full path in the back case of the ternary statement. If you want to make it a little prettier, you can have something like:

```
source: {
    if (root.showCloseIcon) {
        return "/usr/share/icons/.../back.svg";
    } else {
        return "graphics/home.svg"; 
    }
}
```

qml/Launcher/LauncherPanel.qml
```diff
@@ -90,16 +91,21 @@ Rectangle {
...
                 height: width
-                sourceSize.width: width
-                sourceSize.height: height
                 anchors.centerIn: parent
-                source: "graphics/home.svg"
+                source: {
+                    if (root.showCloseIcon) {
+                        return "/usr/share/icons/suru/actions/scalable/back.svg"
+                    }
+                    return "graphics/home.svg"
```
dobey: Please put ending semicolons for statements in JS blocks, for consistency (the rest of the JS added by this diff includes them). :)

dobey:
Alright, it looks good so far. What's the plan for updating the tests? :)

UniversalSuperBox:
Well, I spent several hours trying to get to the point where I could run unity8 on amd64 so I might be able to run the tests. Now that that VM is running I should be able to try my hand at it.

UniversalSuperBox:
Alright, here's what I know so far:

* The tests cannot be run on a headless unit. At the very least, a GL-capable GPU must be available.  We might be able to run tests on our amd64 CI server, but ARM is a no-go. That might be a bug, is xvfb supposed to work?
* A lot of tests broke on the 5.9 transition and were simply disabled.

In order to run the tests, I must start a Mir server (I've become partial to running `miral-kiosk`) then run the tests as `make uitests`.

dobey:
> Alright, here's what I know so far:
> 
> * The tests cannot be run on a headless unit. At the very least, a GL-capable GPU must be available.  We might be able to run tests on our amd64 CI server, but ARM is a no-go. That might be a bug, is xvfb supposed to work?

That sounds like a bug. xvfb-run should definitely work (it works in many other places).

> * A lot of tests broke on the 5.9 transition and were simply disabled.

Disabled how? Seemingly not by the standard test skip methods, as grepping for those doesn't show much.

> In order to run the tests, I must start a Mir server (I've become partial to running `miral-kiosk`) then run the tests as `make uitests`.

You shouldn't need to do that. You should be able to run the `runtests.sh` script in the tree, which runs `make -k xvfballtests`

UniversalSuperBox:
> Disabled how?

https://github.com/ubports/unity8/commit/95d8efbee7f6d3c4ef7f1d091f36d64d2155fd80

> You should be able to run the runtests.sh script in the tree, which runs make -k xvfballtests

Should is the operator there

dobey:
> [95d8efb](https://github.com/ubports/unity8/commit/95d8efbee7f6d3c4ef7f1d091f36d64d2155fd80)

Eww.

UniversalSuperBox:
Alright, got xvfb working now, let's fix some tests: https://github.com/ubports/unity8/commits/xenial_-_edge_-_unity8tests

UniversalSuperBox:
Okay, we should be ready to go.

screenshot20190416_095625904

mariogrip (approved):
LGTM :+1: Great job!

PR #123, opened 2019-04-05
Remove here

mariogrip:
This removes here location thingy


NeoTheThird:
Seems good to me, don't see any errors and the warning on startup related to here service is gone. :) But i'm not confident enough in my knowledge of the stack to vouch for this. So, maybe we could get a review from @dobey?

dobey:
Looks OK to me.

PR #124, opened 2019-04-07
Shell: Show launcher after unlock or if all apps are closed

fredldotme:
Addresses: https://github.com/ubports/ubuntu-touch/issues/1064

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -379,10 +379,30 @@ StyledItem {
         Connections {
             target: greeter
             onActiveChanged: {
-                if (!greeter.active && greeterLoader.openDrawerAfterUnlock) {
+                if (greeter.active)
+                    return
+
+                // Show drawer in case showHome() requests it
+                if (greeterLoader.openDrawerAfterUnlock) {
                     launcher.openDrawer(false);
                     greeterLoader.openDrawerAfterUnlock = false;
                 }
+                // Show the launcher when there are no running apps
+                else if (topLevelSurfaceList.count < 1) {
+                    launcher.switchToNextState("visible")
+                }
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


fredldotme:
Build fail details: seems as if CI failed because `pbuilder-satisfydepends-dummy` dependencies cannot be fulfilled.

NeoTheThird:
Seems to be a problem with our build system (shots fired @mariogrip) that we have to figure out... I pushed your commit to a `xenial_-_edge_-_fredl` and it [builds just fine](https://ci.ubports.com/blue/organizations/jenkins/ubports%2Funity8/detail/xenial_-_edge_-_fredl/1/pipeline/24).

NeoTheThird:
Installed it via ubports-qa and it works fine. Good job!

Just one remark: If you go to the desktop now, the launcher is opened as intended. But if you tap the desktop, it is closed. Is that what we want? Or do we want to replicate the behavior of the "always show launcher" option on the desktop, eg. not hide it when the background is tapped? My opinion is that the behavior from your PR is ok, but maybe others see this differently. :)

cibersheep:
The default behaviour for desktop is that the launcher is always visible. We got the autohide behaviour by default on the phones.
Therefore, I would expect that tapping on desktop hides the Launcher on phones but not on desktop (if that makes sense)

NeoTheThird:
Ok, that seems like a conclusive line of arguments to me. Since that is the behavior we get with this PR, let's go ahead and merge this. Good job @fredldotme! Since you're already meddling with unity 8 papercuts, maybe you'd be interested in taking on [this application spread issue](https://github.com/ubports/ubuntu-touch/issues/1070) next?

NeoTheThird (approved):
lgtm

PR #127, opened 2019-04-18
Stage: 'No running apps' hint on top of Spread

fredldotme:
Initial version until I (or someone else) figure out how to animate the hint opacity/visibility with the spread edge gesture.

Concerns bug report https://github.com/ubports/ubuntu-touch/issues/1070

PR #129, opened 2019-04-19
Blurred user background in app drawer

krille-chan:
screenshot20190418_222017314

A simple eyecandy workaround for the not working blur effect. See here in action: https://janian.de/index.php/s/jqoS9XsdRA2XbY3

This changes are based on ubports:xenial_-_edge_-_sorting

Unfortenetely I don't have the hardware to test this on different sizes...

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -21,6 +21,8 @@ import Utils 0.1
 import "../Components"
 import Qt.labs.settings 1.0
 import GSettings  1.0
+import AccountsService 0.1
+import QtGraphicalEffects 1.0
 
 FocusScope {
     id: root
@@ -72,6 +74,26 @@ FocusScope {
         anchors.fill: parent
         color: "#111111"
         opacity: 0.99
+        
+        Image {
+            id: background
+            anchors.fill: parent
+            source: AccountsService.backgroundFile
+            fillMode: Image.PreserveAspectCrop
+        }
+
+        FastBlur {
+            anchors.fill: background
+            source: background
+            radius: 128
+        }
+
+        // Workaround for images with fastblur can't use opacity
+        Rectangle {
+            anchors.fill: background
+            color: parent.color
+            opacity: 0.67
+        }
 
         MouseArea {
             id: drawerHandle
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -74,7 +74,7 @@ FocusScope {
         anchors.fill: parent
         color: "#111111"
         opacity: 0.99
-        
+
         Image {
             id: background
             anchors.fill: parent


krille-chan:
CI build failed. Looks like the same error like in #127

mariogrip:
Oh! I really like this! :D Will take a look at the ci

mariogrip:
Ok, issue should be fixed now. fixed by: https://github.com/ubports/build-tools/commit/105a02ccc8bfcd60d4b8995bebee58a3f3a8dfd1 

I have triggered a rebuild

krille-chan:
CI still doesn't like me:

```
The following tests FAILED:

	 16 - Whitespace (Failed)

Errors while running CTest

Makefile:130: recipe for target 'test' failed

make[1]: *** [test] Error 8

make[1]: Leaving directory '/build/varlibjenkinsworkspaceubportsunityPRunityubportsgbpbdsc/obj-x86_64-linux-gnu'

dh_auto_test: make -j12 test ARGS+=-j12 returned exit code 2

debian/rules:12: recipe for target 'build-arch' failed

make: *** [build-arch] Error 2

dpkg-buildpackage: error: debian/rules build-arch gave error exit status 2

I: copying local configuration

E: Failed autobuilding of package

I: user script /var/cache/pbuilder/build/cow.1356/tmp/hooks/C10shell starting

  ┌──────────────────────────────────────────────────────────────────────────────┐

  │  FTBFS - problem with building Debian package                                │

  └──────────────────────────────────────────────────────────────────────────────┘

  ┌──────────────────────────────────────────────────────────────────────────────┐

  │  DEB_KEEP_BUILD_ENV is not set to 'true', not keeping build environment      │

  └──────────────────────────────────────────────────────────────────────────────┘

```

UniversalSuperBox:
That's a genuine error in your changes.

```
/build/varlibjenkinsworkspaceubportsunityPRunityubportsgbpbdsc/qml/Launcher/Drawer.qml: bad whitespace in line 77
```

mariogrip:
I changed the base since sorting got merged into edge :)

mariogrip:
But i do wonder, should we make this an setting set option? same as we did for dash background.

krille-chan:
> That's a genuine error in your changes.
> 
> ```
> /build/varlibjenkinsworkspaceubportsunityPRunityubportsgbpbdsc/qml/Launcher/Drawer.qml: bad whitespace in line 77
> ```

I'm sorry for this 😟. I should make indentations visible in Atom...

dobey:
> But i do wonder, should we make this an setting set option? same as we did for dash background.

No we shouldn't. :)

krille-chan:
Hi, are there any news about this? Are there new design decisions where I can help to implement maybe?

dobey:
How does this look in landscape (on a tablet for example), or in windowed mode on an external screen?

krille-chan:
> How does this look in landscape (on a tablet for example), or in windowed mode on an external screen?

I will try to test this later today :)

krille-chan:
Okay I was able to test it with my nexus 5 in convergence mode on a full hd display:
screenshot20190517_113535025
screenshot20190517_113659229

NeoTheThird:
It feels very natural on the M10, great work!

kugiigi:
Looks great on Nexus 7.
Not sure though if making the source image "visible" to false would make sense for the performance.

screenshot20190519_164149660
screenshot20190519_164131680

dobey:
Hrmm. It looks like the wallpaper is not properly aligned in the blur versus the actual wallpaper, which makes things look a bit off, in this case. :-/

NeoTheThird:
It needs an offset of the width of the launcher, it looks like

kugiigi:
To me, it's not really necessary to make it exactly aligned to the wallpaper. The goal is to make the drawer look better and not actually fool people into thinking there's real blurring. It doesn't look too off to me and people will be opening the drawer on top of apps most of the time anyway.

dobey:
> To me, it's not really necessary to make it exactly aligned to the wallpaper. The goal is to make the drawer look better and not actually fool people into thinking there's real blurring. It doesn't look too off to me and people will be opening the drawer on top of apps most of the time anyway.

But it's also a simple fix, and while it might be fine to ignore for a lot of backgrounds, it's still wrong and creates cognitive work for those who notice that something is not quite right, when trying to use the system. It needs to be aligned properly.

dobey (changes requested):
Please fix the alignment issue of the blurred image versus the real background.

krille-chan:
> Please fix the alignment issue of the blurred image versus the real background.

I will try this in the next days

krille-chan:
I have tried it today but I wasn't able to fix this. Don't know how to align an image perfectly to a size, which is not the parent element... Also it seems like that the background of the lock screen and the real desktop background aren't aligned too.

NeoTheThird:
I don't know if the alignment should hold us up right now while there are way bigger problems still. It's not that blatantly annoying with the blur, so i'd suggest filing a bug that someone can look at it someday, but not waste too much time on it now.

dobey:
So, I haven't managed to get horizontal alignment resolved, but I did manage to fix it vertically on my nexus 5. Though it seems the lock screen has a similar alignment issue.

dobey (approved):

dark-eye:
Maybe [this](https://gitlab.com/dark-eye/camera-app/blob/issue_125/OverlayBlur.qml)  can help i made it to the camera-app, i`m currently trying to clean it up a little but it`s probably a kinda working base

PR #135, opened 2019-04-26
Use visibility, not width, to hide the drawer handle

UniversalSuperBox:
This solves a problem where the dots on the handle would be visible even though the handle had 0 width. Seems like a much better solution anyway. I guess I assumed that an invisible component would still take up space, but it doesn't.


cibersheep:
If you add `Opacity: 0` it will not be visible but will take space
`visible: false`, as you say, is not visible and doesn't take the space

dobey (approved):

PR #136, opened 2019-05-01
Replace webbrowser-app with morph-browser

UniversalSuperBox:
About what it says on the tin, replace webbrowser-app with morph-browser where needed. This keeps Morph in the default favorites.

I'm not sure if the changes to the tests are necessary, but this is one case where I wanted to be too liberal at first and then pare down.


mariogrip:
Jeeez, how may tests used this :P

dobey (approved):
Looks fine, though I think it's a bit superfluous in the tests.

UniversalSuperBox:
That's fine, but it keeps everything somewhat in order.

PR #143, opened 2019-05-20
Don't set handleVisible as it's removed.

dobey:
Stop setting the property that was removed to keep the handle always visible.


dobey:
I'm going to go ahead and merge this to fix the build.

PR #144, opened 2019-05-28
Turn the tests back on and fix everything that it breaks

UniversalSuperBox:
1.5 months later...

These commits turn on the Unity8 qmltests and run them automatically during a deb build using dh_auto_test. Then, we fix the myriad race conditions caused by not checking if a component was actually in a state to receive input, or not waiting for the effects of that input to resolve.

This isn't all a bunch of busywork though, it gave me insight into fixing the following bugs with window management:

* Fixes https://github.com/ubports/ubuntu-touch/issues/1081
* Fixes https://github.com/ubports/ubuntu-touch/issues/1124

It will alert us in the future if we do something bad and break some tested behavior. We will start adding tests to confirm and fix bugs in the same PR we fix them.

This does crank up the time to build in CI to an hour. The only way we'll fix this is by optimizing Unity8 itself.

If anyone ever turns the tests off again, they're getting whacked over the head with a Nexus 7.


dobey (approved):
Yay tests!

debian/control
```diff
@@ -72,6 +74,7 @@ Build-Depends: cmake,
                qtdeclarative5-ubuntu-content1,
                qtdeclarative5-ubuntu-settings-components (>= 0.11),
                ttf-ubuntu-font-family,
+               ubuntu-system-settings,
```
mardy: Why is this needed? Are you sure you don't need `libsystemsettings-dev` instead?

qml/ApplicationMenus/MenuItem.qml
```diff
@@ -77,11 +77,11 @@ ActionItem {
...
         id: row
         spacing: units.gu(1)
-        anchors.left: parent.left
-        anchors.right: parent.right
+        anchors.left: root.left
+        anchors.right: root.right
         anchors.leftMargin: units.gu(1)
         anchors.rightMargin: units.gu(1)
-        anchors.verticalCenter: parent.verticalCenter
+        anchors.verticalCenter: root.verticalCenter
```
mardy, line 84: This change (and the two above it) are not needed, and when using anchors it's preferable to use `parent`, than referencing by id. I would revert these.

qml/Launcher/Drawer.qml
```diff
@@ -29,12 +29,85 @@ FocusScope {
...
 
     signal applicationSelected(string appId)
 
+    // Request that the Drawer is opened fully, if it was partially closed then
+    // brought back
+    signal open();
+
+    // Request that the Drawer (and maybe its parent) is hidden, normally if
+    // the Drawer has been dragged away.
+    signal hide();
```
mardy: I would call these `openRequested` and `hideRequested`, so that on the receiving side you'll have `onOpenRequested` and `onHideRequested`. `open()` and `hide()` are good names for functions, not for signals (see also `applicationSelected` above: it's not just `select`).

mardy: Also, remove the `;` at the end of those lines

qml/Launcher/Drawer.qml
```diff
@@ -29,12 +29,85 @@ FocusScope {
...
+    property var oldSelectionStart: null
+    property var oldSelectionEnd: null
+
+    anchors {
+        onRightMarginChanged: {
+            if (fullyOpen && hadFocus) {
+                // See onDraggingHorizontallyChanged below
+                searchField.focus = hadFocus;
+                searchField.select(oldSelectionStart, oldSelectionEnd);
+                return;
```
mardy: you don't need `return`

qml/Shell.qml
```diff
@@ -552,7 +552,7 @@ StyledItem {
             }
 
             readonly property bool focusedSurfaceIsFullscreen: topLevelSurfaceList.focusedWindow
-                ? topLevelSurfaceList.focusedWindow.state === Mir.FullscreenState
+                ? topLevelSurfaceList.focusedWindow.state == Mir.FullscreenState
```
mardy, line 555: This looks very suspicious; if it works with the new code but not with the old, it looks like we might have some error in registering the enums.

qml/Stage/Stage.qml
```diff
@@ -967,34 +967,34 @@ FocusScope {
                         appDelegate.activate();
                     }
                     onStateChanged: {
-                        if (model.window.state === Mir.MinimizedState) {
+                        if (value == Mir.MinimizedState) {
```
mardy, line 970: Same here: there's no reason why the old code wouldn't work, unless we have some serious problem.

qml/Stage/TabletStage.qml
```diff
@@ -775,7 +775,7 @@ AbstractStage {
                         if (priv.mainStageDelegate && stage === ApplicationInfoInterface.SideStage) {
                             return priv.mainStageDelegate.fullscreen;
                         } else if (surface) {
-                            return surface.state === Mir.FullscreenState;
+                            return surface.state == Mir.FullscreenState;
```
mardy, line 778: same here

qml/Stage/WindowedFullscreenPolicy.qml
```diff
@@ -23,7 +23,7 @@ import Unity.Application 0.1
 // state of the window is returned to restored.
 QtObject {
     function applyPolicy(surfaceState, surfaceChrome) {
-        if (surfaceState === Mir.FullscreenState && surfaceChrome === Mir.LowChrome) {
+        if (surfaceState == Mir.FullscreenState && surfaceChrome === Mir.LowChrome) {
```
mardy, line 26: same here

mardy (changes requested):
Nice to see the tests coming back. I've left a few inline comments; nothing super critical, though.

qml/Shell.qml
UniversalSuperBox: I agree there's no reason, but the bug is architecture-specific. Strict comparison worked on amd64 but not armhf or arm64. The values being returned are correct but for whatever reason they don't have the right type.

debian/control
UniversalSuperBox: It was needed for the Wizard tests for app updates, which were then removed by 6653caea530347286883bea94f8d3ef4595aa91c. I'll remove it here.

qml/ApplicationMenus/MenuItem.qml
UniversalSuperBox: Trying to use `parent` caused QML warnings and the items didn't anchor correctly. See 468a1a62144063816dd355a6a715e9fc056a5f81

mardy: what was the warning? Does it go away if you add a binding `parent: root`?

UniversalSuperBox:
```
QWARN  : qmltestrunner::MenuBar::test_stray_submenus() file:///home/dalton/unity8/qml/ApplicationMenus/MenuItem.qml:80: TypeError: Cannot read property of null
QWARN  : qmltestrunner::MenuBar::test_stray_submenus() file:///home/dalton/unity8/qml/ApplicationMenus/MenuItem.qml:81: TypeError: Cannot read property of null
QWARN  : qmltestrunner::MenuBar::test_stray_submenus() file:///home/dalton/unity8/qml/ApplicationMenus/MenuItem.qml:84: TypeError: Cannot read property of null
```

I can log parent in `Component.onCompleted` and get `qml: MenuItem_QMLTYPE_12_QML_18(0x1d43d50, "menuBar-item0-menu-item1-actionItem")` as expected.

This does not go away with a `parent: root` binding.

UniversalSuperBox:
I've fixed some of these issues, @mardy. Could I get another review?

qml/ApplicationMenus/MenuItem.qml
mardy: OK for your changes, but I wonder if we should file a bug on the uitk's ActionItem.

qml/Shell.qml
mardy: Could you please file a bug about this, so when someone has some time he can investigate this issue?

mardy (approved):
Thanks! This looks good to me; I'm still puzzled by the substitution of `===` to `==` when comparing enums; maybe you could file a bug about it, in order to document the problem (and hopefully address it, once someone gets some time to look into it)?

UniversalSuperBox:
https://github.com/ubports/unity8/issues/148

PR #151, opened 2019-06-18
Make sure we don't halt shutdown/reboot wating for apps to close

mariogrip:
This was terrible to begin with, it relied on that the app would close
without problems and got popped from the array.

There is no api for getting all apps, so a hacky workaround for this is
just to check next index number for an app, and repeat this 20 times, if
the apps still wont close, just continue shutdown.

This fixes: https://github.com/ubports/ubuntu-touch/issues/1111

I know, it's ugly! Todo for later is to add an function to get a list of the apps, then use that to loop over. for desktop we should tell the uses of miss behaving apps before rebooting, phone we will just continue the shutdown (maybe? we might want to ask the phone user too, i know its not "normal in android" but it would make sense)

### DIFF
--- a/qml/Components/Dialogs.qml
+++ b/qml/Components/Dialogs.qml
@@ -34,12 +34,21 @@ MouseArea {
     // to be set from outside, useful mostly for testing purposes
     property var unitySessionService: DBusUnitySessionService
     property var closeAllApps: function() {
+        // FIXME! This is an awefull way of doing this!
+        var i = 0;
+        var tries = 0;
         while (true) {
-            var app = ApplicationManager.get(0);
+            var app = ApplicationManager.get(i);
             if (app === null) {
-                break;
+                // if we are at index 0 and app = null, we closed all apps
+                // Try 20 times, if the apps wont close just move on
+                if (i === 0 || tries >= 20)
+                    break;
+                i = 0;
+                tries++;
             }
             ApplicationManager.stopApplication(app.appId);
+            i++;
         }
     }
     property string usageScenario


UniversalSuperBox (changes requested):
I think it'd be better to introduce a bit of code into AppMan which SIGKILLs the app after a few seconds (if a dialog to ask for force closing isn't implemented yet). We definitely aren't crunching right now so we can do it right.

mariogrip:
> I think it'd be better to introduce a bit of code into AppMan which SIGKILLs the app after a few seconds (if a dialog to ask for force closing isn't implemented yet). We definitely aren't crunching right now so we can do it right.

Reason why i didn't want to handle that right now is that we need to change unity-api, qtmir and u8. But its something we need to do yes, but wanted to fix stuff with the current state where canonical left it before starting to bump abi, changing api, adding new functions etc. Change less now, focus on fixing papercuts, not adding enchantments. I see no reason to sigkill tbh, reboot will do that anyway.

qml/Components/Dialogs.qml
```diff
@@ -34,12 +34,21 @@ MouseArea {
     // to be set from outside, useful mostly for testing purposes
     property var unitySessionService: DBusUnitySessionService
     property var closeAllApps: function() {
+        // FIXME! This is an awefull way of doing this!
+        var i = 0;
```
fredldotme, line 38: Shouldn't `i` count down from the length to zero?

mariogrip: there is no way of getting the list or length of the amount of apps in the current api. This just counts up until it hits NULL then resets it, and does this 20 times, if not it just continues reboot/shutdown

UniversalSuperBox: Wouldn't this happen near-instantly on most devices? From what I've seen, Unity8 can ask apps to stop *really* fast.

qml/Components/Dialogs.qml
```diff
@@ -34,12 +34,21 @@ MouseArea {
     // to be set from outside, useful mostly for testing purposes
     property var unitySessionService: DBusUnitySessionService
     property var closeAllApps: function() {
+        // FIXME! This is an awefull way of doing this!
```
dobey, line 37: _awful_ not _awefull_ :)

qml/Components/Dialogs.qml
```diff
@@ -34,12 +34,21 @@ MouseArea {
...
             if (app === null) {
-                break;
+                // if we are at index 0 and app = null, we closed all apps
+                // Try 20 times, if the apps wont close just move on
+                if (i === 0 || tries >= 20)
+                    break;
+                i = 0;
+                tries++;
             }
             ApplicationManager.stopApplication(app.appId);
```
dobey, line 50: Instead of doing all the above, we should probably be checking the return value of `stopApplication` as it returns `true` if the app was stopped, or `false` otherwise. If there is still some issue, and this is returning `true` but unity8 is still hanged in this loop (are we sure that unity8 is hanging in this loop on shutdown anyway?), then we should probably look deeper as the bug is perhaps in qtmir rather than unity8 QML.

dobey (changes requested):

mariogrip:
Closing this, this better: https://github.com/ubports/unity8/pull/159

PR #153, opened 2019-06-19
Fix display of the default wallpaper in the Drawer

### LINKED ISSUE #152: Default wallpaper is not showin in the Drawer
https://github.com/ubports/unity8/pull/129 introduced a blurred user background in the Drawer, but it does not seem to work for the default background. Once a background is set by the user, it is used.

*stub issue because I'm about to submit the PR*

UniversalSuperBox:
It seems that our default setting for AccountsService.backgroundFile
references a file that doesn't exist any more. That's fine since the
default wallpaper is now handled differently in Unity8. However, we
needed to extend that handling to the Drawer.

This commit switches from using AccountsService.backgroundFile as our
wallpaper source to the cached scaled-down wallpaper held by
wallpaperResolver in Shell.qml. Since we're blurring the wallpaper
anyway, we can use this scaled-down version and save some memory.

Fixes #152

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -31,6 +31,7 @@ FocusScope {
     readonly property bool moving: appList && appList.moving
     readonly property Item searchTextField: searchField
     readonly property real delegateWidth: units.gu(10)
+    property url background
     visible: x > -width
     property var fullyOpen: x === 0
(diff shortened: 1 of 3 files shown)


NeoTheThird (approved):
geez, building takes forever. But qml files run fine, so lgtm gj :)

Issue #130, opened 2019-04-21
[edge] Unity8's own focus is not respected


mariogrip:
Unity8's own focus is not respected. 

A good example for this is have an app that shows the keyboard, then drag the top panel down, here it does not give the focus back to unity8 that results in the keyboard not getting dismissed.

UniversalSuperBox:
It should be noted that this also causes the lock screen to not dismiss the keyboard, meaning someone can type in your apps even if the device is locked (assuming the keyboard was available when you locked the device)

mariogrip:
Fixed by: https://github.com/ubports/unity8/pull/156

PR #156, opened 2019-06-26
Make sure no apps have focus if stage is not active/focused

mariogrip:
Make sure no apps have focus if stage is not active/focused, to achive
this a null window is used to steal focus from the other apps.

The reason why a null window is used is that we now use miral to handle
all focusing of apps, and since thats only aware of clients it wont know
if unity8 itself wants the focus. So to workaound this we make a null
window have the focus while unitu8 have focus.

Fixes: https://github.com/ubports/unity8/issues/130
Fixes: https://github.com/ubports/ubuntu-touch/issues/1123
Fixes: https://github.com/ubports/ubuntu-touch/issues/595

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -38,6 +38,7 @@ Q_LOGGING_CATEGORY(TOPLEVELWINDOWMODEL, "toplevelwindowmodel", QtInfoMsg)
 namespace unityapi = unity::shell::application;
 
 TopLevelWindowModel::TopLevelWindowModel()
+    : m_nullWindow(createWindow(nullptr))
 {
 }
 
@@ -718,4 +719,10 @@ void TopLevelWindowModel::activateTopMostWindowWithoutId(int forbiddenId)
             window->activate();
         }
     }
+}
+
+void TopLevelWindowModel::activateNullWindow()
+{
+    m_nullWindow->activate();
+    m_surfaceManager->activate(nullptr);
 }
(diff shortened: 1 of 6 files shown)


NeoTheThird:
[Here](https://paste.ubuntu.com/p/R3WKTmV7kH/)'s a log of this in action on cooler with two browser windows open, switching betwen them and the application randomly, opening and closing the keyboard in all three. At the end of the log, the keyboard no longer showed up in the browser window even though a text field was focused with the cursor blinking.

So there might still be some kinks to iron out here.

mariogrip:
> [Here](https://paste.ubuntu.com/p/R3WKTmV7kH/)'s a log of this in action on cooler with two browser windows open, switching betwen them and the application randomly, opening and closing the keyboard in all three. At the end of the log, the keyboard no longer showed up in the browser window even though a text field was focused with the cursor blinking.
> 
> So there might still be some kinks to iron out here.

I cannot seem to reproduce this. Switched both fast and slow for a good while, could not seem to get a state where the osk no longer showed up.

oh! did you by accident also install qtmir on the same branch? I forgot i added a test to qtmir with the same branch (https://github.com/ubports/qtmir/commits/xenial_-_edge_-_fixfocus) since that will cause some wierdness. I'll go ahead and delete those from the repo now

mariogrip:
oh! wait, if i have a separate app it does indeed get to that state, will debug more :)

mariogrip:
There that should fix the issue @NeoTheThird :)

NeoTheThird (approved):
Seems like that worked. I also don't see https://github.com/ubports/ubuntu-touch/issues/595 any more, so that's an lgtm from me.

Issue #145, opened 2019-05-28
[Edge] Wrong size of the Indicators Bar


cibersheep:
Tested devices: E5 **2019-05-28**, **N5 2019-15/2**, **2019-05-28**
1) Unlock device
2) Place device on landscape
3) Pull indicators bar other than notifications

- Indicators are shifted (sometimes to the right, sometimes to the left).
- Shifting increases by rotating the device sometimes showing shifting also in portrait mode.
- If notification bar is pulled from notifications area (if not empty) looks centered

![imatge](https://user-images.githubusercontent.com/6640041/58499165-d5d75700-817f-11e9-8f5b-41f523af8bae.png)

mariogrip:
Fixed by: https://github.com/ubports/unity8/pull/158

Flohack74:
I still have this on my OPO on edge, from 2019-09-21

Flohack74:
Ok maybe its different, I will make a new ticket...

PR #158, opened 2019-06-28
[panel] Let qt handle calculation of contentX to avoid race condition

mariogrip:
This removes the manual calculation of contentX to avoid race condition
that happen where both us and qt wants to set contentX, this ended up in
a wierd state where it calculated wrong pos for the panel content.

Also as qt states in the docs, we should NOT set contentX ourself:
https://doc.qt.io/qt-5/qml-qtquick-listview.html#positionViewAtIndex-method

This fixes: https://github.com/ubports/unity8/issues/145

### DIFF
--- a/qml/Panel/MenuContent.qml
+++ b/qml/Panel/MenuContent.qml
@@ -45,7 +45,6 @@ Rectangle {
         model: content.model
 
         highlightFollowsCurrentItem: false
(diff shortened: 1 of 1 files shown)


NeoTheThird (approved):
yay for declarative binding loops

PR #159, opened 2019-06-29
Let the windowmanager handle closing of all apps

mariogrip:
This moves the closing of all apps into the window manager where it uses
the right calls to close applications. It also won't block the UI, it does this by sending a signal once all apps are closed. We can later add a popup with the apps that wont close on desktop, but on phone it will just kill them and reboot. (see https://github.com/ubports/unity8/pull/160)

This fixes: https://github.com/ubports/ubuntu-touch/issues/1111

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -436,7 +436,13 @@ void TopLevelWindowModel::removeAt(int index)
         setFocusedWindow(nullptr);
     }
 
-    INFO_MSG << " after " << toString();
+    if (m_closingAllApps) {
+        if (m_windowModel.isEmpty()) {
+            Q_EMIT closedAllWindows();
+        }
+    }
+
+    INFO_MSG << " after " << toString() << " apps left " << m_windowModel.count();
 }
 
 void TopLevelWindowModel::setInputMethodWindow(Window *window)
@@ -725,4 +731,12 @@ void TopLevelWindowModel::activateNullWindow()
 {
     if (!m_nullWindow->focused())
         m_nullWindow->activate();
+}
+
+void TopLevelWindowModel::closeAllWindows()
+{
+    m_closingAllApps = true;
+    for (auto win : m_windowModel) {
+        win.window->close();
+    }
 }
--- a/qml/Components/Dialogs.qml
+++ b/qml/Components/Dialogs.qml
@@ -33,15 +33,6 @@ MouseArea {
 
     // to be set from outside, useful mostly for testing purposes
     property var unitySessionService: DBusUnitySessionService
-    property var closeAllApps: function() {
-        while (true) {
-            var app = ApplicationManager.get(0);
-            if (app === null) {
-                break;
-            }
-            ApplicationManager.stopApplication(app.appId);
-        }
-    }
     property string usageScenario
     property size screenSize: Qt.size(Screen.width, Screen.height)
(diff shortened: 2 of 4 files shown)


UniversalSuperBox:
Is this still needed with https://github.com/ubports/qtmir/pull/26?

UniversalSuperBox:
No longer needed, superseded by https://github.com/ubports/qtmir/pull/26

mariogrip:
This pr is still valid, and should be considered as an replacement for the hacky while(true) loop

plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -725,4 +731,12 @@ void TopLevelWindowModel::activateNullWindow()
...
         m_nullWindow->activate();
+}
+
+void TopLevelWindowModel::closeAllWindows()
+{
+    m_closingAllApps = true;
+    for (auto win : m_windowModel) {
+        win.window->close();
+    }
 }
```
dobey: Please add a newline at end of file.

qml/Components/Dialogs.qml
```diff
@@ -243,9 +245,11 @@ MouseArea {
                 focus: true
                 text: i18n.ctr("Button: Power off the system", "Power off")
                 onClicked: {
-                    root.closeAllApps();
-                    powerDialog.hide();
-                    root.powerOffClicked();
+                    topLevelSurfaceList.closeAllWindows();
+                    doOnClosedAllWindows = function() {
+                        powerDialog.hide();
+                        root.powerOffClicked();
```
dobey, line 251: Please order these calls consistently for each of these `doOnClosedAllWindows` blocks.

qml/Components/Dialogs.qml
```diff
@@ -296,9 +302,11 @@ MouseArea {
         }
 
         onLogoutReady: {
-            root.closeAllApps();
-            Qt.quit();
-            unitySessionService.endSession();
+            topLevelSurfaceList.closeAllWindows();
+            doOnClosedAllWindows = function() {
+                Qt.quit();
```
dobey, line 307: Do we need to actually call `Qt.quit();` here? Shouldn't this happen as a result of the `unitySessionService.endSession();` call instead?

dobey (changes requested):

qml/Components/Dialogs.qml
mariogrip: It's needed as it's a cleaner shutdown then a call from upstart. Also this makes it work without an unity session.

qml/Components/Dialogs.qml
mariogrip: These are left how the order they was in.

plugins/WindowManager/TopLevelWindowModel.cpp
mariogrip: why do we need this? it makes it really inconvenience when vscode removes this automatically anyway...

UniversalSuperBox: In Code, enable `Files: Insert Final Newline` and `Files: Trim Final Newlines`. This will ensure there is exactly one newline character at the end of each file.

As for why... https://stackoverflow.com/a/729795

mariogrip: Thanks :) will update it :)

mariogrip:
Had to rebase, should compile now

dobey (approved):

PR #164, opened 2019-07-02
Add multi monitor and workspace support

mariogrip:
This adds multi monitor and workspace support. Workspace is disabled by default, but can be forced on.

This is massive!

Needs:
https://github.com/ubports/qtmir/pull/25
https://github.com/ubports/unity-api/pull/15


dobey:
This is massive, I'm not sure it's the right direction to take unity8 in, and we don't have functional parity with pre-app-drawer unity8 in edge yet, so it seems very early to be thinking about it at all, really. :)

mariogrip:
> This is massive, I'm not sure it's the right direction to take unity8 in, and we don't have functional parity with pre-app-drawer unity8 in edge yet, so it seems very early to be thinking about it at all, really. :)

True, and i agree with you fully, i said the same 2 weeks ago. But after looking at the changes done to enable miral, it will be some work for us to enable convergence without this as this brings in some breaking changes in that area. So my idea here is to "try", see what breaks and what works good. 

It also makes the conflicts easier to fix later when we do them side by side with development :)

So with that said, there is no saying this will be merged in "the mir/u8 OTA" at all.

mariogrip:
On the plus side it does seem to work well with convergence on the nexus5

UniversalSuperBox:
Please request review when this is ready for it.

mariogrip:
Ok tests are fixed now. Had two i had to disable for now, but those are not deal breakers for now (as these are for menybar).

mariogrip:
pr-merge is expected to fail, as it does not have qtmirserver-dev.

NeoTheThird:
Branch is failing too.

mariogrip:
Disabled the segfaulting test for now, doing this so we can test it on devices in while this is being fixed.  

**DO NOT MERGE UNTIL THIS IS FIXED**

PR #165, opened 2019-07-11
Show the greeter when mainApp goes away

dobey:
In case of dialer-app crashing or being killed in a call, when no
other apps are open, the greeter needs to be shown again, so the
phone can be unlocked.


UniversalSuperBox (changes requested):
We've found that this is better solved by handling the situation in `_onMainAppChanged` as that is where the Greeter is notified of the app changing.

PR #167, opened 2019-07-11
Make Shell tests way faster

UniversalSuperBox:
The search for dashCommunicator was taking 5-7 seconds per run of loadShell. Removing this improves test run time by 12 minutes locally (which should end up being about 1/6th of build time in CI).


UniversalSuperBox:
Woo! 43 minutes versus 62 minutes!

dobey (approved):

Issue #170, opened 2019-06-21
Fullscreen apps are remembered, will not go back to not-fullscreen


NeoTheThird:
The browser is showing some unexpected behavior on edge, it seems to be running in full-screen. This leads to the following issues:

- The indicator bar is not shown
- The browser can not be moved in a window in desktop mode
- The launcher and app drawer can not be opened in desktop mode
- The top bar shows up while dragging out the launcher and drawer and disappears again when the the drawer is fully extended, leading to weird shifting around
- Working with multiple windows is a pain in the ass
- Side stage sometimes works and sometimes doesn't
- Top bar disappears if browser is focused in side-stage and re-appears if focus is switched to the app in main stage

I don't see anything interesting going on in the logs for browser or unity8. Might be caused by ubports/unity8#144. Have not observed this behavior with other full-screen apps such as Pathwind.

UniversalSuperBox:
It seems like this sort of problem affects any app which uses fullscreen, so I've moved it to unity8.

UniversalSuperBox:
Fixed by https://github.com/ubports/unity8/pull/172

PR #172, opened 2019-08-02
[Stage] Make sure we only use saved states when appropriate

mariogrip:
If the app is fullscreen at startup, we should not use saved state
Example of why: if you open game that only requests fullscreen at
Statup, this will automaticly be set to "restored state" since thats the
default value of stateStorage, this will result in the app having the
"restored state" as it will not make a fullscreen call after the app has
started.

If saved state is fullscreen, we should use app inital state
Example of why: if you open browser with youtube video at fullscreen and
close this app, it will be fullscreen next time you open the app.

This fixes: https://github.com/ubports/ubuntu-touch/issues/1081
This fixes: https://github.com/ubports/unity8/issues/170

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -1021,6 +1021,25 @@ FocusScope {
                 onWindowReadyChanged: {
                     if (windowReady) {
                         var loadedMirState = WindowStateStorage.toMirState(windowStateSaver.loadedState);
+                        var state = loadedMirState;
+
+                        if (window.state == Mir.FullscreenState) {
+                            // If the app is fullscreen at startup, we should not use saved state
+                            // Example of why: if you open game that only requests fullscreen at
+                            // Statup, this will automaticly be set to "restored state" since
+                            // thats the default value of stateStorage, this will result in the app
+                            // having the "restored state" as it will not make a fullscreen
+                            // call after the app has started.
+                            console.log("Inital window state is fullscreen, not using saved state.");
+                            state = window.state;
+                        } else if (loadedMirState == Mir.FullscreenState) {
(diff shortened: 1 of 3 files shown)


UniversalSuperBox:
Wow. That was really complicated.

It's stylistic, but might it be better to not *store* the Fullscreen State rather than not *restore* it? Keep the data clean, keep the output clean.

UniversalSuperBox:
Whichever way you prefer, please add QML regression tests to ensure this fix behaves as you expect. It appears that there's a test failing on arm64, too

mariogrip:
tbh, i got no clue why arm64 fails...

```
FAIL!  : qmltestrunner::PhoneStage::test_select(4) property focusedApplicationId
   Actual   (): unity8-dash
   Expected (): dialer-app
   Loc: [/build/unity8-8.20+ubports+0~20190802235447.1~1.gbp3d71fd/tests/qmltests/Stage/tst_PhoneStage.qml(317)]
```

But i added tests for my changed at least.

mariogrip:
Now two is failing on arm64... makes no sense since my changes should not effect focusedApplicationId

mariogrip:
Wait, it seems like it might be 
```
Mesa 11.2.0 implementation error: Failed to link fixed function fragment shader: error: count of uniform locations > MAX_UNIFORM_LOCATIONS(4294967295 > 98304)

Please report at https://bugs.freedesktop.org/enter_bug.cgi?product=Mesa
Segmentation fault
```
that seems to also stop the test early, and probably have something to do with all the fails too.

mariogrip:
Issue was wrong libwayland version so mesa got conflicted. fixed and builds fine now

dobey (approved):
Looks ok.

PR #174, opened 2019-08-14
Add watchers for xdg desktop file changes

### LINKED ISSUE #1058: Drawer does not refresh apps
In Unity8 on Edge, the app Drawer is used to launch apps rather than the App Scope. Currently, the Drawer does not refresh when an app is installed or removed and stays in an invalid state. There is also no UI interaction to refresh the Drawer (such as pulling down and releasing).

The Drawer should refresh automatically when apps are installed or removed.

mariogrip:
This adds watchers for xdg desktop file changes. It also figures out if
the appid id is "ubuntu-app-launch" or "standard xdg" style.

This spawns the first code to replace most (if not all) of
ubuntu-app-launch, but with QT!

This fixes: https://github.com/ubports/ubuntu-touch/issues/1058

### DIFF
--- a/plugins/Unity/Launcher/appdrawermodel.cpp
+++ b/plugins/Unity/Launcher/appdrawermodel.cpp
@@ -16,13 +16,15 @@
 
 #include "appdrawermodel.h"
 #include "ualwrapper.h"
+#include "xdgwatcher.h"
 
 #include <QDebug>
 #include <QDateTime>
 
 AppDrawerModel::AppDrawerModel(QObject *parent):
     AppDrawerModelInterface(parent),
-    m_ual(new UalWrapper(this))
+    m_ual(new UalWrapper(this)),
+    m_xdgWatcher(new XdgWatcher(this))
 {
     Q_FOREACH (const QString &appId, UalWrapper::installedApps()) {
         UalWrapper::AppInfo info = UalWrapper::getApplicationInfo(appId);
@@ -37,9 +39,9 @@ AppDrawerModel::AppDrawerModel(QObject *parent):
     }
 
     // keep this a queued connection as it's coming from another thread.
-    connect(m_ual, &UalWrapper::appAdded, this, &AppDrawerModel::appAdded, Qt::QueuedConnection);
-    connect(m_ual, &UalWrapper::appRemoved, this, &AppDrawerModel::appRemoved, Qt::QueuedConnection);
-    connect(m_ual, &UalWrapper::appInfoChanged, this, &AppDrawerModel::appInfoChanged, Qt::QueuedConnection);
+    connect(m_xdgWatcher, &XdgWatcher::appAdded, this, &AppDrawerModel::appAdded, Qt::QueuedConnection);
+    connect(m_xdgWatcher, &XdgWatcher::appRemoved, this, &AppDrawerModel::appRemoved, Qt::QueuedConnection);
+    connect(m_xdgWatcher, &XdgWatcher::appInfoChanged, this, &AppDrawerModel::appInfoChanged, Qt::QueuedConnection);
 }
 
 int AppDrawerModel::rowCount(const QModelIndex &parent) const
--- a/plugins/Unity/Launcher/xdgwatcher.cpp
+++ b/plugins/Unity/Launcher/xdgwatcher.cpp
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2019 UBports Foundation
+ * Author(s): Marius Gripsgard <marius@ubports.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "xdgwatcher.h"
+
+#include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QStandardPaths>
+#include <QTextStream>
+
+XdgWatcher::XdgWatcher(QObject* parent)
+    : QObject(parent),
+      m_watcher(new QFileSystemWatcher(this))
+{
+    connect(m_watcher, &QFileSystemWatcher::directoryChanged, this, &XdgWatcher::onDirectoryChanged);
+    connect(m_watcher, &QFileSystemWatcher::fileChanged, this, &XdgWatcher::onFileChanged);
+
+    const auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
+    for (const auto &path: paths) {
+        const auto qdir = QDir(path);
+        if (!qdir.exists()) {
+            continue;
+        }
+
+        // Add the path itself to watch for newly added apps
+        m_watcher->addPath(path);
+
+        // Add watcher for eatch app to watch for changes
+        const auto files = qdir.entryInfoList(QDir::Files);
+        for (const auto &file: files) {
+            if (file.suffix() == "desktop") {
+                const auto path = file.absoluteFilePath();
+                m_watcher->addPath(path);
+                m_registry.insert(path, getAppId(file));
+            }
+        }
+    }
+}
+
+// "Ubuntu style" appID is filename without versionNumber after last "_"
+const QString XdgWatcher::stripAppIdVersion(const QString rawAppID) const {
+    auto appIdComponents = rawAppID.split("_");
+    appIdComponents.removeLast();
+    return appIdComponents.join("_");
+}
+
+// Standard appID see:
+// https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#desktop-file-id
+const QString XdgWatcher::toStandardAppId(const QFileInfo fileInfo) const {
+    const auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
+    for (const auto &path: paths) {
+        if (fileInfo.absolutePath() == path) {
+            break;
+        }
+        if (fileInfo.absolutePath().contains(path)) {
+            auto fileStr = fileInfo.absoluteFilePath();
+            fileStr.replace(path, "");
+            fileStr.replace("/", "-");
+            fileStr.replace(".desktop", "");
+            return fileStr;
+        }
+    }
+    return fileInfo.completeBaseName();
+}
+
+const QString XdgWatcher::getAppId(const QFileInfo fileInfo) const {
+    // We need to open the file to check if its and Ual application
+    // as we cant just rely on the app name as "normal" apps can also
+    // contain 3 _ causing us to belive its an ubuntu app
+    // Example kde_org_kate would become kde_org
+    QFile qFile(fileInfo.absoluteFilePath());
+    qFile.open(QIODevice::ReadOnly);
+    QTextStream fileStream(&qFile);
+    QString line;
+    while (fileStream.readLineInto(&line)) {
+        if (line.startsWith("X-Ubuntu-Application-ID=")) {
+            auto rawAppID = line.replace("X-Ubuntu-Application-ID=", "");
+            qFile.close();
+            return stripAppIdVersion(rawAppID);
+        }
+    }
+    qFile.close();
+
+    // If it's not an "Ubuntu" appID, we follow freedesktop standard
+    return toStandardAppId(fileInfo);
+}
+
+// Watch for newly added apps
+void XdgWatcher::onDirectoryChanged(const QString &path) {
+    const auto files = QDir(path).entryInfoList(QDir::Files);
+    const auto watchedFiles = m_watcher->files();
+    for (const auto &file: files) {
+        const auto appPath = file.absoluteFilePath();
+        if (file.suffix() == "desktop" && !watchedFiles.contains(appPath)) {
+            m_watcher->addPath(appPath);
+
+            const auto appId = getAppId(file);
+            m_registry.insert(appPath, appId);
+            Q_EMIT appAdded(appId);
+        }
+    }
+}
+
+void XdgWatcher::onFileChanged(const QString &path) {
+    QFileInfo file(path);
+    if (file.exists()) {
+        // The file exists, this must be an modify event
+        Q_EMIT appInfoChanged(m_registry.value(path));
+    } else {
+        // File does not exist, assume this is an remove event.
+        // onDirectoryChanged will handle rename event
+        Q_EMIT appRemoved(m_registry.take(path));
+    }
+}
--- a/plugins/Unity/Launcher/xdgwatcher.h
+++ b/plugins/Unity/Launcher/xdgwatcher.h
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2016 Canonical, Ltd.
+ * Copyright (C) 2019 UBports Foundation
+ * Author(s): Marius Gripsgard <marius@ubports.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -14,30 +15,32 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <QFileSystemWatcher>
+#include <QHash>
+#include <QFileInfo>
 
-#include <unity/shell/launcher/AppDrawerModelInterface.h>
+class QFileInfo;
 
-#include "launcheritem.h"
-
-class UalWrapper;
-class XdgWatcher;
-
-class AppDrawerModel: public AppDrawerModelInterface
+class XdgWatcher: public QObject
 {
     Q_OBJECT
 public:
-    AppDrawerModel(QObject* parent = nullptr);
+    XdgWatcher(QObject* parent = nullptr);
 
-    int rowCount(const QModelIndex &parent) const override;
-    QVariant data(const QModelIndex &index, int role) const override;
-
-private Q_SLOTS:
+Q_SIGNALS:
     void appAdded(const QString &appId);
     void appRemoved(const QString &appId);
     void appInfoChanged(const QString &appId);
 
+private Q_SLOTS:
+    void onDirectoryChanged(const QString &path);
+    void onFileChanged(const QString &path);
+
 private:
-    QList<LauncherItem*> m_list;
-    UalWrapper *m_ual;
-    XdgWatcher *m_xdgWatcher;
+    const QString toStandardAppId(const QFileInfo fileInfo) const;
+    const QString getAppId(const QFileInfo file) const;
+    const QString stripAppIdVersion(const QString rawAppID) const;
+
+    QFileSystemWatcher* m_watcher;
+    QHash<const QString, QString> m_registry;
 };
--- a/tests/plugins/Unity/Launcher/xdgwatcher.cpp
+++ b/tests/plugins/Unity/Launcher/xdgwatcher.cpp
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2016 Canonical, Ltd.
+ * Copyright (C) 2019 UBports Foundation.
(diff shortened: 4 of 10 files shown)


plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,117 @@
...
+    : QObject(parent),
+      m_watcher(new QFileSystemWatcher(this))
+{
+    connect(m_watcher, &QFileSystemWatcher::directoryChanged, this, &XdgWatcher::onDirectoryChanged);
+    connect(m_watcher, &QFileSystemWatcher::fileChanged, this, &XdgWatcher::onFileChanged);
+
+    auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
+    for (auto &path: paths) {
+        auto qdir = QDir(path);
+        if (!qdir.exists())
```
dobey: Please consistently use open/closing curly braces around single-line if statements.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,117 @@
...
+    auto appIdComponents = rawAppID.split("_");
+    appIdComponents.removeLast();
+    return appIdComponents.join("_");
+}
+
+const QString XdgWatcher::toStandardAppId(const QFileInfo fileInfo) const {
+    auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
+    for (auto &path: paths) {
+        if (fileInfo.absolutePath() == path)
+            break;
```
dobey, line 68: Single line if statement inconsistency.

plugins/Unity/Launcher/xdgwatcher.h
```diff
@@ -0,0 +1,43 @@
...
+    void onDirectoryChanged(const QString &path);
+    void onFileChanged(const QString &path);
+
+private:
+    const QString toUbuntuAppId(const QString rawAppID) const;
+    const QString toStandardAppId(const QFileInfo fileInfo) const;
+    const QString toAppId(const QFileInfo file) const;
+
+    QFileSystemWatcher* m_watcher;
+};
```
dobey: No newline at end of file.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (file.exists()) {
+        // The file exists, this must be an modify event
+        auto appId = toAppId(file);
+        Q_EMIT appInfoChanged(appId);
+    } else {
+        // File does not exist, assume this is an remove event.
+        // onDirectoryChanged will handle rename event
+        Q_EMIT appRemoved(toStandardAppId(file), toUbuntuAppId(file.completeBaseName()));
+    }
+}
```
dobey: No newline at end of file.

tests/plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+void XdgWatcher::addMockApp(const QString &appId)
+{
+    Q_EMIT appAdded(appId);
+}
+
+void XdgWatcher::removeMockApp(const QString &appId)
+{
+    Q_EMIT appRemoved(appId, "");
+}
```
dobey: Add newline at end of file.

tests/plugins/Unity/Launcher/xdgwatcher.h
```diff
@@ -0,0 +1,38 @@
...
+    void removeMockApp(const QString &appId);
+Q_SIGNALS:
+    void appAdded(const QString &appId);
+    void appRemoved(const QString &appId, const QString &ubuntuAppId);
+    void appInfoChanged(const QString &appId);
+
+private:
+    static QStringList s_list;
+    static XdgWatcher *s_xinstance;
+};
```
dobey: Add newline at end of file.

dobey (changes requested):

mariogrip:
@dobey Should be fixed up now :)

plugins/Unity/Launcher/appdrawermodel.cpp
```diff
@@ -82,11 +84,11 @@ void AppDrawerModel::appAdded(const QString &appId)
     endInsertRows();
 }
 
-void AppDrawerModel::appRemoved(const QString &appId)
+void AppDrawerModel::appRemoved(const QString &appId, const QString &ubuntuAppId)
```
mardy: Can you please add a comment to explain the difference between the `appId` and `ubuntuAppId`?

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,119 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "xdgwatcher.h"
+#include <QStandardPaths>
+#include <QDebug>
+#include <QDir>
+#include <QFileInfo>
+#include <QFile>
+#include <QTextStream>
```
mardy, line 24: Please add a space after the `xdgwatcher.h` include, and sort the other ones alphabetically (nitpicking, I know :-) )

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+XdgWatcher::XdgWatcher(QObject* parent)
+    : QObject(parent),
+      m_watcher(new QFileSystemWatcher(this))
+{
+    connect(m_watcher, &QFileSystemWatcher::directoryChanged, this, &XdgWatcher::onDirectoryChanged);
+    connect(m_watcher, &QFileSystemWatcher::fileChanged, this, &XdgWatcher::onFileChanged);
+
+    auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
```
mardy: add `const` before `auto` (also in a few places below)

plugins/Unity/Launcher/appdrawermodel.cpp
mardy: Or maybe rename the second variable to `unversionedAppId`?

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,119 @@
...
+        auto qdir = QDir(path);
+        if (!qdir.exists()) {
+            continue;
+        }
+
+        // Add the path itself to watch for newly added apps
+        m_watcher->addPath(path);
+
+        // Add watcher for eatch app to watch for changes
+        for (auto &file: qdir.entryInfoList(QDir::Files)) {
```
mardy: move the call to `entryInfoList()` outside of the `for` loop, and save the result into a const variable.

Reason: "When used with a (non-const) object that has copy-on-write semantics, the range-based for loop may trigger a deep copy by (implicitly) calling the non-const begin() member function. If that is undesirable (for instance because the loop is not actually modifying the object), std::as_const can be used"

mardy (changes requested):

mariogrip:
@mardy There we go, that should fix it :) Also found a little error in my toStandardAppId() thats fixed now.

mardy (approved):

plugins/Unity/Launcher/appdrawermodel.cpp
```diff
@@ -82,11 +84,12 @@ void AppDrawerModel::appAdded(const QString &appId)
     endInsertRows();
 }
 
-void AppDrawerModel::appRemoved(const QString &appId)
+// ubuntuAppId is appID without version suffix, see xdgwatcher.cpp.
+void AppDrawerModel::appRemoved(const QString &appId, const QString &ubuntuAppId)
```
dobey: Instead of passing both in here, why not make the functions (which seem to not require the class object itself to be useful), into public static methods, and then just call `XdgWatcher::stripAppIdVersion(appid)` when needed below? This avoids a semi-duplicated argument, and makes the code easier to follow.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+        for (const auto &file: files) {
+            if (file.suffix() == "desktop") {
+                m_watcher->addPath(file.absoluteFilePath());
+            }
+        }
+    }
+}
+
+// Ubuntu appID is filename without versionNumber after last "_"
+const QString XdgWatcher::toUbuntuAppId(const QString rawAppID) const {
```
dobey: I think this should be renamed to not have "Ubuntu" in the name. Perhaps it should be `stripAppIdVersion` instead.

plugins/Unity/Launcher/appdrawermodel.cpp
```diff
@@ -82,11 +84,12 @@ void AppDrawerModel::appAdded(const QString &appId)
...
 {
     int idx = -1;
     for (int i = 0; i < m_list.count(); i++) {
-        if (m_list.at(i)->appId() == appId) {
+        if (m_list.at(i)->appId() == appId || m_list.at(i)->appId() == ubuntuAppId) {
```
dobey: Do we even need the second comparison here (and thus the argument or need to strip the version) at all? Shouldn't the `appId` that gets passed in already have the version stripped always? In what cases would it not be?

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (C) 2019 UBports project.
```
dobey: If copyright is being assigned to the foundation, I think we should use the Foundation's legal name here (and anywhere else it's used like this).

mariogrip: Cant use that yet as foundation is still not full up.

plugins/Unity/Launcher/appdrawermodel.cpp
mariogrip: The reason for this is to make sure we don't strip an "standard" appID by accident. Example is kate_kde.desktop, we would strip "_kde" in that case. Since the file has been removed we have no way of knowing if it's a "ubuntu style" app id.

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Yeah make sense, will update this

plugins/Unity/Launcher/appdrawermodel.cpp
mariogrip: problem with that is that the "normal/standard" appID might not be the same as strippedAppId as the "ubuntu style" app id does not add the parent folder into app id.

example:
.local/share/desktop/kde/kate.desktop would be kde-kate as appID.

mariogrip: example where these conflics:

.local/share/desktop/ubports/system-settings_ubports_1.1.1.desktop would then be passed as ubports-system-settings_ubports_1.1.1 to appRemoved, then stripped to ubports-system-settings_ubports and that would be wrong, as "ubuntu style" appID does not append "folder name" as xdg one does

dobey: But, click hook does not create desktop files in a subdir, and we get the app ID for these from inside the desktop file, so this isn't an issue. I don't see any need to have extra special handling in here. The model has a list of app IDs, and one gets removed. there doesn't need to be extra logic for special cases in here.

plugins/Unity/Launcher/appdrawermodel.cpp
dobey: OK, I re-read the comment in XdgWatcher, and this is a hack to deal with the fact the file was removed. I will have to think a bit on this, but I'm certain we can handle it a better way.

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: I thought legal status had been granted?

dobey (changes requested):
Because previous comments and GitHub is broke.

UniversalSuperBox (changes requested):
https://ci.ubports.com/blue/organizations/jenkins/ubports%2Funity8/detail/xenial_-_edge_-_appwatcher/4/pipeline#step-38-log-192

mariogrip:
@dobey that should be better, now its a static method that gets called instead of passing it with the signal.

mariogrip:
Cool! It builds now. Just ignore the others as they are only here due to changes to Jenkins.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+                m_watcher->addPath(file.absoluteFilePath());
+            }
+        }
+    }
+}
+
+// "Ubuntu style" appID is filename without versionNumber after last "_"
+const QString XdgWatcher::stripAppIdVersion(const QString rawAppID) {
+    auto appIdComponents = rawAppID.split("_");
+    appIdComponents.removeLast();
```
dobey, line 58: I think we should only do this, if the `appIdComponents.length() == 3`, otherwise it is not a valid versioned app ID for the purposes here. This would also solve the concern above that we might strip `kde_kate` improperly. It would also simplify the code a fair bit, as we could then avoid special-casing below, and above, only passing the result of `toAppId()` around.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+            fileStr.replace(".desktop", "");
+            return fileStr;
+        }
+    }
+    return fileInfo.completeBaseName();
+}
+
+const QString XdgWatcher::toAppId(const QFileInfo fileInfo) const {
+    QFile qFile(fileInfo.absoluteFilePath());
+    qFile.open(QIODevice::ReadOnly);
```
dobey, line 87: Actually, I think we can just get rid of the whole need to open the file and check for `X-Ubuntu-Application-ID` entirely, and only use `toStandardAppId()` to get the app ID, since click hooks place the files as the correct `${APP_ID}.desktop` filename. Then we can just always strip the version.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+        // The file exists, this must be an modify event
+        const auto appId = toAppId(file);
+        Q_EMIT appInfoChanged(appId);
+    } else {
+        // File does not exist, assume this is an remove event.
+        // onDirectoryChanged will handle rename event
+
+        // As we have no way of checking the deleted file, we need to try remove both types
+        // of appIDs.
+        Q_EMIT appRemoved(toStandardAppId(file));
```
dobey: Won't this trigger a remove event for all click packages, causing an odd race condition scenario where the entry may be removed from the app drawer, even though the app is still installed?

dobey (changes requested):
A few more concerns that came up thinking about this.

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: no, since click will remove then create the desktop file. Since there will always be one remove and one add event. also the remove event will always pick the one first in the array/map (the oldest one).  while add will add to the bottom of the array/map.

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Yeah that make sense, will add that :)

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Yeah i guess that should also work just fine, since we use to remove the app anyway.

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: So, "UBports project" is not a legal entity which can have Copyright assigned to it. It either needs to be assigned to `UBports Foundation` or otherwise should say `Marius Gripsgard` given you are the author and thus retain the copyright (unless you've copied code from elsewhere that is under a suitable license where you can release it as GPLv3).

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: This has still not been updated?

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: This has still not been updated?

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: Hmm. OK.

plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <QTextStream>
+
+XdgWatcher::XdgWatcher(QObject* parent)
+    : QObject(parent),
+      m_watcher(new QFileSystemWatcher(this))
+{
+    connect(m_watcher, &QFileSystemWatcher::directoryChanged, this, &XdgWatcher::onDirectoryChanged);
+    connect(m_watcher, &QFileSystemWatcher::fileChanged, this, &XdgWatcher::onFileChanged);
+
+    const auto paths = QStandardPaths::standardLocations(QStandardPaths::ApplicationsLocation);
```
dobey, line 33: As this will only watch directories in the host, this PR doesn't solve the issue for install/remove of apps in Libertine. How are we going to handle those?

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Thinking about this, we may still improperly strip for example `org_kde_kate`, so we would still need to open the file to check

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Thinking about this, we may still improperly strip for example org_kde_kate, so we would still need to open the file to check

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: So the idea to that was to make symlinks the .desktop to the standard locations instead, this makes it easier for libertine work on other DE (if needed)

mariogrip:
So, I added an internal registry, this way we will always give the correct id. This also really simplifies the code overall. and I think i covered all your comments @dobey. Please re-review :)

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Now that the Foundation is up, i  edited this.

tests/plugins/Unity/Launcher/xdgwatcher.cpp
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (C) 2019 UBports project.
```
dobey: Should be _Foundation_ instead.

tests/plugins/Unity/Launcher/xdgwatcher.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (C) 2019 UBports project.
```
dobey: Should be _Foundation_ instead.

plugins/Unity/Launcher/xdgwatcher.cpp
dobey: I don't think this is a real concern. KDE does not install applications in this manner.

dobey (changes requested):

tests/plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: wops, forgot those files

plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: kde was just an example, but one app that does do this on my local system is `krita_jpg_jpeg.desktop`

tests/plugins/Unity/Launcher/xdgwatcher.cpp
mariogrip: Fixed

dobey (approved):
I still think this is not quite right, but I don't want to hold it up any longer on semantics.

PR #176, opened 2019-08-15
[WindowManager] Focus surface on startup only if focus is requested

mariogrip:
This focuses newly created surface only if focus is requested by the
user.

Needs: https://github.com/ubports/qtmir/pull/27

Fixes: https://github.com/ubports/ubuntu-touch/issues/1088

### DIFF
--- a/plugins/WindowManager/Window.cpp
+++ b/plugins/WindowManager/Window.cpp
@@ -27,6 +27,7 @@ namespace unityapi = unity::shell::application;
 Q_LOGGING_CATEGORY(UNITY_WINDOW, "unity.window", QtWarningMsg)
 
 #define DEBUG_MSG qCDebug(UNITY_WINDOW).nospace() << qPrintable(toString()) << "::" << __func__
+#define WARNING_MSG qCWarning(UNITY_WINDOW).nospace() << qPrintable(toString()) << "::" << __func__
 
 Window::Window(int id, QObject *parent)
     : QObject(parent)
@@ -184,7 +185,15 @@ void Window::setSurface(unityapi::MirSurfaceInterface *surface)
             setSurface(nullptr);
         });
 
+        // Surface should never be focused at this point!
+        if (m_surface->focused()) {
+            WARNING_MSG << "Inital surface is focused!";
+        }
+
         // bring it up to speed
+        if (m_focused) {
+            m_surface->activate();
+        }
         if (m_positionRequested) {
             m_surface->setRequestedPosition(m_requestedPosition);
         }


plugins/WindowManager/Window.cpp
```diff
@@ -184,7 +184,13 @@ void Window::setSurface(unityapi::MirSurfaceInterface *surface)
             setSurface(nullptr);
         });
 
+        // Surface should never be focused at this point!
+        Q_ASSERT(!m_surface->focused());
```
UniversalSuperBox: If this assertion fails, Unity crashes, correct?

dobey: If debug is enabled (which it should be), then yes, it does a `qFatal()` which is of course, fatal.

dobey (changes requested):
Don't use `Q_ASSERT()`. If the goal is to return from the method when a condition is not met, you should use `if (!condition) { return; }` (appropriately formatted) instead.

plugins/WindowManager/Window.cpp
mariogrip: Only if debug is enabled (as @dobey says). 

but else then debug mode, if this happens in production mode we should let the code go, since it will run updateFocused later where our state will be updated.

mariogrip:
> Don't use `Q_ASSERT()`. If the goal is to return from the method when a condition is not met, you should use `if (!condition) { return; }` (appropriately formatted) instead.

The goal here is not to stop or return, we need to continue if this is built with release binaries. It's not the end of the world, so we can simply ignore it (as updateFocus will happen later in that code). But for development this should break. Q_ASSERT is the perfect fit for this usecase.

dobey:
> > Don't use `Q_ASSERT()`. If the goal is to return from the method when a condition is not met, you should use `if (!condition) { return; }` (appropriately formatted) instead.
> 
> The goal here is not to stop or return, we need to continue if this is built with release binaries. It's not the end of the world, so we can simply ignore it (as updateFocus will happen later in that code). But for development this should break. Q_ASSERT is the perfect fit for this usecase.

The way in which debs are built, is that they are built with debug enabled, and then symbols are stripped and collected into debug packages, so that crashes on user systems can be debugged more easily. Therefore, assertions will result in fatal action. Really, we should never use asserts in runtime code. Anything where it might seem like an assert would make sense, should be replaced by something which will fail through gracefully. Having the shell crash is not graceful at all.

Ironically, you also just commented in another issue about a `Q_ASSERT` being hit at runtime, causing crashing on shutdown. This is exactly why we shouldn't use them. :)

plugins/WindowManager/Window.cpp
dobey: If the idea is we should continue on here, but we want some logging, then I think it'd be better to use a `qCritical()` inside an `if` block.

mariogrip:
> > > Don't use `Q_ASSERT()`. If the goal is to return from the method when a condition is not met, you should use `if (!condition) { return; }` (appropriately formatted) instead.
> > 
> > 
> > The goal here is not to stop or return, we need to continue if this is built with release binaries. It's not the end of the world, so we can simply ignore it (as updateFocus will happen later in that code). But for development this should break. Q_ASSERT is the perfect fit for this usecase.
> 
> The way in which debs are built, is that they are built with debug enabled, and then symbols are stripped and collected into debug packages, so that crashes on user systems can be debugged more easily. Therefore, assertions will result in fatal action. Really, we should never use asserts in runtime code. Anything where it might seem like an assert would make sense, should be replaced by something which will fail through gracefully. Having the shell crash is not graceful at all.
> 
> Ironically, you also just commented in another issue about a `Q_ASSERT` being hit at runtime, causing crashing on shutdown. This is exactly why we shouldn't use them. :)

Ok, changed it to a qWarning. As it's not that critical if it happens. 

oh! that might clears up some stuff indeed. but I still do find it weird, what makes that Q_ASSERT does not print when packaged with debian?

UniversalSuperBox:
All Debian packages are built in debug mode (`-DCMAKE_BUILD_TYPE=Debug`), then the debug symbols are stripped out and put into the automated debug packages (`package-dbgsym.ddeb`).

dobey (approved):
Looks ok now with a warning message.

mariogrip:
@UniversalSuperBox @dobey  Wait no, debian packages are built with "QT_NO_DEBUG" so Q_ASSERT would have been fine. but does not matter now when its merged.

but see: https://ci.ubports.com/blue/rest/organizations/jenkins/pipelines/ubports/pipelines/unity8/branches/xenial_-_edge_-_dontautofocus/runs/3/nodes/26/steps/37/log/?start=0

PR #177, opened 2019-08-16
Let TopLevelWindowModel decide what app to refocus, if any

mariogrip:
This let TopLevelWindowModel deside what app it should refocus after a
the null window has been focused. It will only refocus the previous
window if null focus still has focus. This makes sure we dont give the
focus to another app while an app is starting, or other window has
clamed the focus.

Pending Activation will block refocus of previous focused window
this is needed since a surface activation with miral is async
and a activation of placeholder is sync. This causes a race condition
between placeholder and prev window. This results in prev window
gets focused as it will always be later then placeholder.

This fixes: https://github.com/ubports/ubuntu-touch/issues/1170

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -650,16 +650,19 @@ void TopLevelWindowModel::setFocusedWindow(Window *window)
     if (window != m_focusedWindow) {
         INFO_MSG << "(" << window << ")";
 
-        Window* previousWindow = m_focusedWindow;
+        m_previousWindow = m_focusedWindow;
 
         m_focusedWindow = window;
         Q_EMIT focusedWindowChanged(m_focusedWindow);
 
-        if (previousWindow && previousWindow->focused() && !previousWindow->surface()) {
+        if (m_previousWindow && m_previousWindow->focused() && !m_previousWindow->surface()) {
             // do it ourselves. miral doesn't know about this window
-            previousWindow->setFocused(false);
+            m_previousWindow->setFocused(false);
         }
     }
+
+    // Reset
+    m_pendingActivation = false;
 }
 
 unityapi::MirSurfaceInterface* TopLevelWindowModel::inputMethodSurface() const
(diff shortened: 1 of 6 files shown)


plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -721,8 +724,29 @@ void TopLevelWindowModel::activateTopMostWindowWithoutId(int forbiddenId)
     }
 }
 
-void TopLevelWindowModel::activateNullWindow()
+void TopLevelWindowModel::rootFocus(bool focus)
+{
+    if (focus) {
+        // Give focus back to previous focused window, only if null window is focused
+        // If null window is not focused, a different app had taken the focus and we
+        // should repect that or if a pendigActivation is going on
```
dobey: Typo `pendigActivation` here. And missing punctuation (periods). Along with an extra space in `focused,  a different`.

plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -721,8 +724,29 @@ void TopLevelWindowModel::activateTopMostWindowWithoutId(int forbiddenId)
...
+            m_nullWindow->activate();
+        }
+    }
+}
+
+// Pending Activation will block refocus of previous focused window
+// this is needed since a surface activation with miral is async
+// and a activation of placeholder is sync. This causes a race condition
+// between placeholder and prev window. This results in prev window
+// gets focused as it will always be later then placeholder.
```
dobey: Some grammar fixes needed in this comment block.

`since a surface activation is async` -> `since surface activation is async,`
`and a activation of placeholder` -> `and activation of the placeholder`
`gets focused as` -> `getting focused, as`
`later then` -> `later than`

plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -721,8 +724,29 @@ void TopLevelWindowModel::activateTopMostWindowWithoutId(int forbiddenId)
...
+void TopLevelWindowModel::pendingActivation()
 {
-    if (!m_nullWindow->focused())
-        m_nullWindow->activate();
+    m_pendingActivation = true;
 }
```
dobey: Missing newline at end of file.

qml/Stage/Stage.qml
```diff
@@ -91,19 +91,13 @@ FocusScope {
 
     onInteractiveChanged: {
         // Stage must have focus before activating windows, including null
-        if (interactive) {
-            focus = true;
+        if (interactive) focus = true;
```
dobey: Better I think to leave the code here as was, and just close off the if statement early with closing brace on a new line.

qml/Stage/Stage.qml
```diff
@@ -91,19 +91,13 @@ FocusScope {
...
-            }
-        } else {
-            // Activate null window once stage is not active this makes
-            // sure none of the apps have focus when stage is not active
-            topLevelSurfaceList.activateNullWindow();
-        }
+        // Activate focused app once stage is active again
+        // this makes sure the focused app have focus
+        // Activate null window once stage is not active this makes
+        // sure none of the apps have focus when stage is not active
```
dobey: Some grammar and punctuation fixes needed here too. :)

dobey (changes requested):

mariogrip:
@dobey That should do it :)

dobey (commented):
Could perhaps use some test coverage of this to ensure it doesn't regress in future?

mariogrip:
Ok added tests, lets see how this goes :)

mariogrip:
Cool! It builds now. Just ignore the others as they are only here due to changes to Jenkins.

dobey (approved):

PR #181, opened 2019-09-04
Manage apps on edge (ish)

NeoTheThird:
This kind of fixes https://github.com/ubports/ubuntu-touch/issues/1199 by opening the openstore url, except it doesn't, because of https://gitlab.com/theopenstore/openstore-meta/issues/248. Please wait with merging until that is resolved.

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -292,7 +292,12 @@ FocusScope {
                 readonly property bool focused: index === GridView.view.currentIndex && GridView.view.activeFocus
 
                 onClicked: root.applicationSelected(model.appId)
-                onPressAndHold: Qt.openUrlExternally("https://open-store.io/app/" + model.appId.split("_")[0] + "/")
+                onPressAndHold: {
+                  if (model.appId.includes(".")) { // Open OpenStore page if app is a click
+                    var splitAppId = model.appId.split("_");
+                    Qt.openUrlExternally("https://open-store.io/app/" + model.appId.replace("_" + splitAppId[splitAppId.length-1],"") + "/");
+                  }
+                }
                 z: loader.active ? 1 : 0
 
                 Column {
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -292,6 +292,7 @@ FocusScope {
                 readonly property bool focused: index === GridView.view.currentIndex && GridView.view.activeFocus
 
                 onClicked: root.applicationSelected(model.appId)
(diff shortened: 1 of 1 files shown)


qml/Launcher/Drawer.qml
```diff
@@ -292,6 +292,7 @@ FocusScope {
                 readonly property bool focused: index === GridView.view.currentIndex && GridView.view.activeFocus
 
                 onClicked: root.applicationSelected(model.appId)
+                onPressAndHold: Qt.openUrlExternally("https://open-store.io/app/" + model.appId.split("_")[0] + "/")
```
dobey: I'm not sure we should just open a link in another app like this, directly from `pressAndHold`.  I think we need more integration between Launcher and Drawer in terms of what we do when interacting with app icons.

NeoTheThird: Agreed. But maybe this could be a quick-and-dirty fix for now to make the drawer less broken. 100% with you that drawer and launcher should be more integrated, but given that it takes some refactoring to pull out the quicklistmenu and we don't know what direction UAL will take, i'd put a pin in that. And when the time comes, it's just one line of code to remove.

NeoTheThird:
As Brian pointed out, it was the trailing slash that broke the parsing in the openstore. Fixed now.

qml/Launcher/Drawer.qml
dobey: Hmm. This will also open the store in a broken manner I think, for apps which are not clicks (from debs or libertine), or for locally installed clicks which aren't in the store.

NeoTheThird:
I don't think we have to add tests for this right now, btw. I would wait with that until we refactor the launcher and drawer to use the contextmenu everywhere.

UniversalSuperBox (changes requested):
This isn't a great perfect solution to the problem, but it should be good enough for a first iteration. It will not open the store for anything but click apps now.

But I disagree, it should be tested. We should ensure that it does not try to launch the store for deb-type apps and does for click-type. Could be done by sending a signal to open the store instead, the signal handler doing the actual `Qt.openUrlExternally`, then creating a SignalSpy to watch for that in the tests.

NeoTheThird:
>Could be done by sending a signal to open the store instead, the signal handler doing the actual Qt.openUrlExternally, then creating a SignalSpy to watch for that in the tests.

Wouldn't it be easier to simulate the long-click in the test and check if the openstore is the focused app after a short delay? Or am i missing something?

UniversalSuperBox:
The OpenStore isn't installed in the tests.

NeoTheThird:
Riiiiiiiight. Damn.

UniversalSuperBox (approved):
The functionality here works. It's not the best experience, but it's a better one. Good enough for a first pass.

Issue #132, opened 2019-04-22
Confusing Animation When Pulling down Indicator


cibersheep:
Pull from the indicator bar from Notifications is kind of ok, but pulling it from for example Sound or Battery indicators creates a weird animation that is confusing.

NeoTheThird:
What do you mean exactly? I don't see anything weird going on.

cibersheep:
It might be a N5 thing but if you pull from, let's say, the clock on non edge channels, the drop down is smooth and keep focus on the clock. If you do on edge channel the animation of the panel goes a bit crazy for a moment and then focus is a bit lost.

Can I forward mir to record the screen to the computer somehow? That would make more sense

NeoTheThird:
http://docs.ubports.com/en/latest/userguide/advanceduse/screencasting.html

cibersheep:
Pulling down from wifi indicator (good animation)
https://youtu.be/FYzrf6QjlNs

Pulling from audio indicator (confusing animation)
https://youtu.be/pWSDqM2NETc

cibersheep:
I'm testing this on N5 (on landscape is more obvious).
Recently got an E5 on edge as well. Animation looks correct on E5 edge

mariogrip:
Can you check if this is fixed by: https://github.com/ubports/unity8/pull/158?

cibersheep:
Ouh! Thanks @mariogrip I can test tomorrow / Monday

cibersheep:
Tested on N5 with latest updates and it's not fixed yet.
- Pulling from Notifications, GPS (if visible), Bluetooth (if visible), Network and time indicators works ok
- Pulling from Sound or battery creates this weird animation

If all this indicators are visible: 
- pulling from sound works fine
- Battery creates weird animation

![imatge](https://user-images.githubusercontent.com/6640041/60463631-cc448180-9c4c-11e9-95da-3166fbf124ff.png)

cibersheep:
I have tested on E5 and works good. It looks to be a specific thing for N5?

UniversalSuperBox:
Maybe it's related to screen size. I see it less on the Oneplus One than the FP2 or Nexus 5, but it still happens.

[...]

PR #186, opened 2019-09-05
Fix indicator text binding loop

NeoTheThird:
This fixes a binding-loop in the indicators. Don't know if the loop actually caused any problems, but the noisy logging annoyed the crap out of me. Effects on #132 have to be evaluated, but since i could not reproduce it before, i am not sure if this helps.


dobey (approved):

NeoTheThird:
Thanks :)

PR #187, opened 2019-09-05
Address Showable errors

NeoTheThird:
This is probably very low-priority. The code has been in place since 2013, and i don't know if it actually caused any issues. But... Errors. Must fix errors. So here we go.

When Unity8 is starting, the following logs:
```
[2019-09-05:20:48:19.165] file:///usr/share/unity8//Components/Showable.qml:43: Error: Cannot assign to non-existent property "target"
[2019-09-05:20:48:19.166] file:///usr/share/unity8//Components/Showable.qml:42: Error: Cannot assign to non-existent property "target"
[2019-09-05:20:48:19.199] file:///usr/share/unity8//Components/Showable.qml:43: Error: Cannot assign to non-existent property "target"
[2019-09-05:20:48:19.205] file:///usr/share/unity8//Components/Showable.qml:42: Error: Cannot assign to non-existent property "target"
[2019-09-05:20:48:20.317] file:///usr/share/unity8//Components/Showable.qml:42: Error: Cannot assign to non-existent property "target"
[2019-09-05:20:48:20.317] file:///usr/share/unity8//Components/Showable.qml:43: Error: Cannot assign to non-existent property "target"
```

This PR fixes that without breaking any functionality (afaics), but it causes many many many tests to fail. Could this be because the test assertions are wrong, or did i actually break something here?

### DIFF
--- a/qml/Components/Showable.qml
+++ b/qml/Components/Showable.qml
@@ -39,8 +39,8 @@ Item {
 
     // automatically set the target on showAnimation and hideAnimation to be the
     // showable itself
-    onShowAnimationChanged: if (showAnimation) Object.defineProperty(showAnimation, "target", { value: showable, writable: true });
-    onHideAnimationChanged: if (hideAnimation) Object.defineProperty(hideAnimation, "target", { value: showable, writable: true });
+    onShowAnimationChanged: if (showAnimation && !showAnimation.target) Object.defineProperty(showAnimation, "target", { value: showable, writable: true });
+    onHideAnimationChanged: if (hideAnimation && !hideAnimation.target) Object.defineProperty(hideAnimation, "target", { value: showable, writable: true });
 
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
I think the change is subtle, but important:

`FAIL!  : qmltestrunner::Showable::test_showNow() Showable should be at end of animation after showNow`

I don't think Showables are showing... like at all. Something is quite wrong at least. For example, the Greeter is broken in the test area. It just doesn't show up now:

There is no greeter here

Compared to a checkout of xenial_-_edge:

image

Remind me to write up how I set up a VM for Unity8 development and run these demos, please. I think they'll help everyone in the future.

NeoTheThird:
That's really weird. In production, this just works. Maybe we have to look at the test setup code to figure out if something weird is going on.

But again, low prio.

PR #188, opened 2019-09-08
Welcome page should not have back button

mariogrip:


UniversalSuperBox (approved):

UniversalSuperBox:
I'm pretty sure there was an issue filed about this, but I'm having trouble finding it. Something about being able to put the wizard to a white page by hitting "Back"

PR #189, opened 2019-09-08
Specify QPA platform to mirserver if we use wayland as client QPA

mariogrip:


NeoTheThird:
Don't see a problem with this. Is this still needed? If so, let's get it merged. If not, let's close the PR. Clean the board, people! :)

data/unity8.conf
```diff
@@ -82,6 +82,12 @@ pre-start script
         initctl set-env --global UBUNTU_PLATFORM_API_BACKEND=test.so.3
         initctl set-env --global UBUNTU_PLATFORM_API_SENSOR_TEST=/etc/ubuntusensors/test.sensors
     fi
+
+    # Specify QPA platform to mirserver if we use wayland as client QPA
```
dobey: Is this really necessary? Shouldn't this be changed at https://github.com/ubports/unity8/blob/xenial_-_edge/src/main.cpp#L31 instead? I think we definitely shouldn't be setting the value with `initctl set-env` as that will make all processes trying to use mirserver by default (such as maliit-server), which won't work.

dobey (changes requested):

data/unity8.conf
mariogrip: initctl set-env will just set for unity8, using --global however will set it for all. This is what we use on the pinephone, and it works great. 

Sure we can set it in main.cpp, but i dint want to hardcode it. but i guess it will be fine to hardcode it.

dobey: Either way, I think we shouldn't do this sort of thing in the upstart job file. It makes it more annoying when we will have to support systemd, and for people trying to get unity8 "ported" to other distros where they might use yet some other init systems.

mariogrip:
Should be ok for review now :)

dobey (approved):

prog66:
Hi on OPO it does not work anymore ... the applications Libertine works not libreoffice for example worked more now since October 25 for me.

Issue #185, opened 2019-09-05
Greeter labels out of place


NeoTheThird:
There labels in the greeter (coll: lock-screen) have been out of place for a while now, and it's kind of annoying. It's not so recognizable on the phone (narrow greeter), but on tablets such as my trusty old cooler (wide greeter) it just looks stupid. This is obviously not device-dependent, though; enabling window-mode will let you see this on every device.

### Steps to reproduce (cooler):

1. Unlock the device
2. Press the power-button to lock
3. Press the power-button again to activate the screen
4. Dismiss the keyboard

Notice the out-of-place "Ubuntu" label that does not shift with the rest of the box.

### Logs

Here's what unity8 is logging for the step 2 described above:

```
[2019-09-05:19:42:45.732] qtmir.sensor: OrientationSensor::onDisplayPowerStateChanged
[2019-09-05:19:42:46.051] qtmir.surfaces: MirSurface[0x17442b0,"maliit-server"]::updateState(restored)
[2019-09-05:19:42:46.052] qtmir.surfaces: MirSurface[0x17442b0,"maliit-server"]::updateExposure(true)
[2019-09-05:19:42:46.057] file:///usr/share/unity8//Greeter/GreeterPrompt.qml:239: TypeError: Cannot read property of null
[2019-09-05:19:42:46.070] file:///usr/share/unity8//Greeter/GreeterPrompt.qml:240: TypeError: Cannot read property of null
[2019-09-05:19:42:46.071] file:///usr/share/unity8//Greeter/GreeterPrompt.qml:241: TypeError: Cannot read property of null
[2019-09-05:19:42:46.072] file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Components/1.3/TextField.qml:889: TypeError: Cannot read property of null
[2019-09-05:19:42:46.073] file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Components/1.3/TextField.qml:891: TypeError: Cannot read property of null
[2019-09-05:19:42:46.073] file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Components/1.3/TextField.qml:916: TypeError: Cannot read property of null
[2019-09-05:19:42:46.074] file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Components/1.3/TextField.qml:918: TypeError: Cannot read property of null
[...]

PR #190, opened 2019-09-12
Work around greeter anchoring mess

NeoTheThird:
This hides the user manager to work around #185. Instead, it will display a new label that's properly anchored.

Yes. I know: It's not pretty. But there's a lot of unused complexity here that we might need later. I don't want to touch all that now, but i am also convinced that we need to fix this. *Every* person i give my M10 to notices this, and that is not the first impression i want people to get of Ubuntu Touch. At some point we will have to refactor this and fix it properly.

### DIFF
--- a/qml/Greeter/GreeterPrompt.qml
+++ b/qml/Greeter/GreeterPrompt.qml
@@ -214,14 +214,15 @@ FocusScope {
             id: hint
             objectName: "promptHint"
             anchors {
-                left: parent.left
-                right: parent.right
-                verticalCenter: parent.verticalCenter
+                left: parent ? parent.left  : undefined
+                right: parent ? parent.right  : undefined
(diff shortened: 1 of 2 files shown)


NeoTheThird:
If we decide this is the way to go for now, we can also cherry-pick that into xenial to have it in ota-11. Thoughts?

NeoTheThird:
@UniversalSuperBox

UniversalSuperBox (approved):
This isn't perfect, but it works. Multi-user support in full-greeter seems to be a long way off, maybe never manifesting.

NeoTheThird:
Yep. Plus, multi-user support doesn't get more broken by this, since the label is conditional. I know it isn't pretty (i kinda hate myself), but for now i think it's the right step. Thanks for the merge.

PR #191, opened 2019-09-15
Fix tutorial test and make sure totorial get skipped if spread has been shown

mariogrip:
This:
 - This fixes tutorial test by making verifying synchronous to avoid interfering with touchFlick function.
 - Make sure right tutorial gets skipped if spread has been shown, this also fixes flakiness with other tests.

### DIFF
--- a/qml/Tutorial/TutorialRight.qml
+++ b/qml/Tutorial/TutorialRight.qml
@@ -24,7 +24,7 @@ TutorialPage {
     property string usageScenario
 
     // When on phone or tablet, fade out as the drag progresses
-    opacityOverride: usageScenario === "desktop" ? 1 : 1 - stage.rightEdgeDragProgress * 2
+    opacityOverride: stage.spreadShown ? 0 : usageScenario === "desktop" ? 1 : 1 - stage.rightEdgeDragProgress * 2
 
     Connections {
         target: stage
--- a/tests/qmltests/Tutorial/tst_Tutorial.qml
+++ b/tests/qmltests/Tutorial/tst_Tutorial.qml
@@ -586,8 +586,6 @@ Rectangle {
         }
 
         function test_tutorialRightShortDrag() {
-            skip("This test interferes with test_tutorialRightAutoSkipped. If" +
-                 " one runs before the other, the second one will fail. FIXME");
             var tutorial = findChild(shell, "tutorial");
             var tutorialRight = findChild(tutorial, "tutorialRight");
             var stage = findChild(shell, "stage");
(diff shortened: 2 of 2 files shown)


qml/Tutorial/TutorialRight.qml
```diff
@@ -24,7 +24,7 @@ TutorialPage {
     property string usageScenario
 
     // When on phone or tablet, fade out as the drag progresses
-    opacityOverride: usageScenario === "desktop" ? 1 : 1 - stage.rightEdgeDragProgress * 2
+    opacityOverride: stage.spreadShown ? 0 : usageScenario === "desktop" ? 1 : 1 - stage.rightEdgeDragProgress * 2
```
UniversalSuperBox, line 27: This only hides the tutorial, it doesn't skip it. You'll want to find a way to mark the edge as finished instead.

Also, nested ternaries are super ugly. Would not recommend.

UniversalSuperBox (changes requested):

qml/Tutorial/TutorialRight.qml
mariogrip: no it does mark it as finished, look: https://github.com/ubports/unity8/blob/xenial_-_edge/qml/Tutorial/TutorialPage.qml#L49 that calls hide() that then triggers the animation and triggers the onStopped https://github.com/ubports/unity8/blob/xenial_-_edge/qml/Tutorial/TutorialPage.qml#L83

mariogrip: Also, nested ternaries does not really matter as the logic would be the same ether way if you use if or whatnot. also: https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340

UniversalSuperBox: Okay, I see what's going on now. Try the following test:

```
        function test_rightTutorialSkippedBySpread() {
            // Ensures that the right tutorial is skipped by opening the Spread,
            // but no others.
            AccountsService.demoEdgesCompleted = ["left"];
            var tutorialLeft = findChild(shell, "tutorialLeft");
            var tutorialLeftTimer = findInvisibleChild(tutorialLeft, "tutorialLeftTimer");
            tutorialLeftTimer.interval = 99999;
            AccountsService.demoEdgesCompleted = [];
            compare(AccountsService.demoEdgesCompleted, []);
            wait(2000);
            ApplicationManager.startApplication("gallery-app");
            ApplicationManager.startApplication("facebook-webapp");
            ApplicationManager.startApplication("camera-app");
            touchFlick(shell, shell.width, halfHeight, 0, halfHeight);

            compare(AccountsService.demoEdgesCompleted, ["right"])
        }
```

This fails, `demoEdgesCompleted` appears to be `["left","right"]` with the single flick. We're skipping both the left and right tutorial at the same time.

mariogrip: but this is unrelated to my changed, seems to be some cache issues or what not. but that's a scenario that wont happen, we will never end up in this case.

mariogrip:
Now edge fails to build, it needs this :) This is a fix for THAT :D

```
FAIL!  : qmltestrunner::Tutorial::test_tutorialRightAutoSkipped() property demoEdgesCompleted
   Actual   (): left
   Expected (): [left,right]
   Loc: [/build/varlibjenkinsworkspaceubportsunityxenialedgeunityubportsgbpdcedsc/tests/qmltests/Tutorial/tst_Tutorial.qml(634)]
```

https://ci.ubports.com/blue/rest/organizations/jenkins/pipelines/ubports/pipelines/unity8/branches/xenial_-_edge/runs/48/nodes/26/steps/34/log/?start=0

UniversalSuperBox (approved):
Right, looks like that occurs either way.

PR #192, opened 2019-09-18
Cache the blur results and tone down the radius

UniversalSuperBox:
Since the source of the blur will very rarely change, it can be cached
for faster rendering at the expense of RAM.

https://doc.qt.io/qt-5.9/qml-qtgraphicaleffects-fastblur.html#cached-prop
states that "Visual quality of the blur is reduced when radius exceeds
value 64."
For that reason, I've turned down the absolute blur value.

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -157,7 +157,8 @@ FocusScope {
         FastBlur {
             anchors.fill: background
             source: background
-            radius: 128
+            radius: 64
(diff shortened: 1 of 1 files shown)


dobey (approved):
Not sure why the unrelated test failed, but this looks OK to me.

UniversalSuperBox:
Hmm. We'll have to see if that test starts failing consistently when Unity8 is built like this. I wonder if builds are conflicting.

PR #194, opened 2019-09-19
Fix greeter label positioning

NeoTheThird:
This is #190 for xenial, so we can land a fix for #185 in ota-11.


mariogrip (approved):
LGTM :+1:

UniversalSuperBox (approved):

PR #199, opened 2019-09-28
Enable titleLabel in portrait mode too.

giiba:
This is meant as a design suggestion.

I like the window title that shows up when using my phone in landscape and wanted it to display all the time. I've been using the Panel this way for a couple months and have noticed no ill effects. I will upload some photos after posting this.

Perhaps it could also be an option in the settings menu of the phone, but I do not know how to implement such.


giiba:
Some screenshots to demonstrate:

screenshot20190928_130021663
 
screenshot20190928_130234338
 
 
It could free app designers to use title space for something else:
screenshot20190928_130316428

dobey:
This would presumably place the title of the focused window in the top panel, in windowed mode as well, which would be wrong. In most apps (at least that use Ubuntu.Components MainView) it creates redundant information, and also implies one should be able to swipe down in order to get the app's "menu" (which will work for legacy apps in edge perhaps, but not for converged apps without menus).

giiba:
That is a use case I never encounter as I only use my phone in staged mode. I will see what happens in windowed mode when I get a chance.

kugiigi:
Can you also try what happens when you have a lot of icons in the indicator? My MX4 can fill all the width in portrait like when I enable bluetooth, cellular, and wifi at the same time.

dobey:
> Can you also try what happens when you have a lot of icons in the indicator? My MX4 can fill all the width in portrait like when I enable bluetooth, cellular, and wifi at the same time.

Right. This will be even more an issue on devices like Nexus 4 and E4.5 that have smaller screens.

UniversalSuperBox:
Closing for now, I think we've found that there is not enough space in the top bar to have the title label showing.

PR #202, opened 2019-10-01
Make the appDelegate know when it's being focused

UniversalSuperBox:
There's a "slide" animation that plays when apps are focused on their
own request or on the request of an outside party.

Unfortunately, this animation plays after the rightEdgeFocusAnimation,
which calls 'appDelegate.activate()' when it stops. When that happens,
the delegate acts like it was focused by another source and plays the
sliding animation. This behavior was probably introduced in f1dcc66.

The inhibitSlideAnimation property of each appDelegate gets set when it
is focused with the flip animation and unset when it becomes unset. This
strikes a good balance between simplicity and ensuring all the
functionality (sliding in and out when another app is focused, for
example) still works.

Also, ensure the slide animations take the speed of the animations
imposed by the Stage.

Unfortunately, it's not possible to regression test this with the current test framework. The problem behavior doesn't even occur in the tests. I suspect it's QtMir-only, then, but QtMir isn't exactly behaving badly in this case... It isn't asked to focus the window until after the animations complete.

Fixes https://github.com/ubports/ubuntu-touch/issues/1121

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -829,6 +829,13 @@ FocusScope {
                     onHeightChanged: appDelegate.updateNormalGeometry();
                 }
 
+                // True when the Stage is focusing this app and playing its own animation.
+                // Stays true until the app is unfocused.
+                // If it is, we don't want to play the slide in/out transition from StageMaths.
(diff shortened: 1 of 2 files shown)


mariogrip (approved):
works and LGTM :+1:

PR #203, opened 2019-10-03
Reject quick drags on the panel menu

UniversalSuperBox:
Very quick drags could cause the panel to call out to launch the dialer during a phone call, which is generally not helpful when you're trying to drag the indicators down quickly.

Fixes https://github.com/ubports/ubuntu-touch/issues/1198


mariogrip (approved):
Works fine and LGTM :+1:

PR #205, opened 2019-10-14
Don't allow the user to snipe indicator handles

UniversalSuperBox:
Prior to these changes, the user could grab the ShowDragHandle while the PanelMenu was hiding. This caused strange behavior.

Fixes https://github.com/ubports/ubuntu-touch/issues/1257


mariogrip (approved):
LGTM :+1:

PR #206, opened 2019-10-15
Hide keyboard when Launcher or Drawer are dragged

UniversalSuperBox:
The Launcher and Drawer do not respect the on-screen keyboard rectangle, so about half of them are hidden when the Drawer's search field is focused.

To fix this, unfocus the Drawer's search field when either of them move.

This does not remove focus when the user scrolls with the mouse wheel, only when they click and drag or touch and drag.

Fixes https://github.com/ubports/ubuntu-touch/issues/1238
Fixes https://github.com/ubports/ubuntu-touch/issues/1245

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -115,6 +115,10 @@ FocusScope {
         searchField.focus = true;
     }
 
+    function unFocusInput() {
+        searchField.focus = false;
+    }
+
     Keys.onPressed: {
         if (event.text.trim() !== "") {
             focusInput();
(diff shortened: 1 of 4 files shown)


dobey (approved):
Nice, and with tests. ♥

mariogrip (approved):
LGTM :+1:

Danfro:
Sorry for the late reply.

Just tested on my E5 on dev, but it was the same on edge until a few days ago before I changed to dev:

When scrolling the launcher the OSK does NOT disappear. The issue regarding the bottom half of the launcher not being accessible therefore is NOT fixed.

Issue #211, opened 2019-10-22
Drawer: Cannot clear search field after search


lduboeuf:
Edge, N5: Type a search in the search bar, then try to touch the "close" (x) button, nothing happen.

Expected result: search field is cleared, full app list is shown. 

Note that a touch in the search bar, clear the field

UniversalSuperBox:
Fixed by #213

PR #213, opened 2019-10-23
Bug fixes on Drawer (fixes #212, #211)

### LINKED ISSUE #212: Drawer: search make UI freeze
Edge, N5: Type an app search, keyboard key overlay stay visible for about a second, app list is refreshed but we can see the lag

( Test case: > 100 apps installed )

lduboeuf:
#212 : There was a dependency issue with the model AppDrawerProxyModel bound to GridView , the source pointed to  sortProxyModel (AppDrawerProxyModel) which in turn points to AppDrawerModel.
That structure leads to search be fired multiple times. That cause the lags in search result and UI freeze.

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -275,11 +275,7 @@ FocusScope {
                 height: rows * delegateHeight
                 clip: true
 
-                model: AppDrawerProxyModel {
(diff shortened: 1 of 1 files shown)


UniversalSuperBox (approved):
Ah, much better.

PR #216, opened 2019-10-24
provide a default icon for apps (fixes #210)

### LINKED ISSUE #210: default icon for an app
In case of resource load failure, can we provide a default icon as it was the case on Stable ?

lduboeuf:


dobey (changes requested):
Hmm, I don't think we should copy the image into here, if it's just going to be the same image that's already in the suru theme. Instead we should just use the version from the theme. Also, I think we should change the image so that it doesn't include the Ubuntu logo. It also might be better to solve this issue in the `ubuntu-app-launch` code, which is where icon paths are actually discovered.

lduboeuf:
alright, so better close that PR .

lduboeuf:
so by providing a default icon path in ubuntu-app-launch when anything wrong happens ?

lduboeuf:
https://github.com/ubports/ubuntu-app-launch/blob/xenial/libubuntu-app-launch/application-icon-finder.cpp
here ?

dobey:
Yes. By falling back to the default app icon from the theme, if no icon path is found.

lduboeuf:
> Yes. By falling back to the default app icon from the theme, if no icon path is found.

ok, i see. Well i'm not able to build/test ubuntu-app-launch so can't do it.
With crossbuilder it does not work :-/ like others project.

PR #226, opened 2019-11-09
Add battery state led indicators.

wdehoog:
Using already existing code from ubports the led is now also used to
indicate battery states full/charging/low. Unread notifications remain highest
priority.

Monitoring battery level and charger state is done in two ways:
- as in system-settings\plugins\battery\PageComponent.qml. (dbus signals
  from com.canonical.indicator.power)
- using unity Wizard Status plugin (dbus signals from
  org.freedesktop.UPower)

The led itself is still controlled using the unity Lights plugin.

### DIFF
--- a/qml/Panel/Indicators/IndicatorsLight.qml
+++ b/qml/Panel/Indicators/IndicatorsLight.qml
@@ -99,30 +99,23 @@ QtObject {
             return
         }
 
-        // show charging. show full only when charging
         var isCharging = batteryIconName.indexOf("charging") >= 0
                          || deviceState != "discharging"
 
-        var isFull = isCharging
-                     && (batteryIconName.indexOf("full-charged") >= 0
-                         || deviceState == "fully-charged"
-                         || batteryLevel >= 100)
+        var isFull = batteryIconName.indexOf("full-charged") >= 0
+                     || deviceState == "fully-charged"
+                     || batteryLevel >= 100
 
-        if (isCharging) {
-            if (isFull)
-                indicatorState = "BATTERY_FULL"
-            else
-                indicatorState = "BATTERY_CHARGING"
-            return
-        }
-
-        // show battery low or nothing
-        if (batteryIconName.indexOf("caution") >= 0
-            || batteryIconName.indexOf("empty") >= 0)
+        if (isCharging && isFull) {
+            indicatorState = "BATTERY_FULL"
+        } else if (isCharging) {
+            indicatorState = "BATTERY_CHARGING"
+        } else if (batteryIconName.indexOf("caution") >= 0
+                   || batteryIconName.indexOf("empty") >= 0) {
             indicatorState = "BATTERY_LOW"
-        else
+        } else {
             indicatorState = "INDICATOR_OFF"
-
+        }
     }
 
     onIndicatorStateChanged: {
--- a/qml/Panel/Indicators/IndicatorsLight.qml
+++ b/qml/Panel/Indicators/IndicatorsLight.qml
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014 Canonical Ltd.
+ * Copyright 2019 UBports Foundation
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


qml/Panel/Indicators/IndicatorsLight.qml
```diff
@@ -22,12 +23,126 @@ import Powerd 0.1
...
+        updateLightState("onBatteryIconNameChanged")
+    }
+
+    function updateLightState(msg) {
+        console.log("updateLightState: " + msg + ", hasMessages: " + hasMessages + ", icon: " 
+            + batteryIconName + ", displayStatus: " + displayStatus + ", deviceState: " 
+            + deviceState + ", batteryLevel: " + batteryLevel)
+
+        // only show led when display is off
+        if(displayStatus == Powerd.On) {
```
mardy: Please follow the existing coding style and add a space after every `if`.

qml/Panel/Indicators/IndicatorsLight.qml
```diff
@@ -22,12 +23,126 @@ import Powerd 0.1
...
+        //   Battery low warning dialog on screen shows up at 10%
+        //   but 'caution' icon at 9%.
+        // 
+
+        var lColor = ""
+        var lOnMS = -1
+        var lOffMS = -1
+
+        var isCharging = batteryIconName.indexOf("charging") >= 0
+        if(!isCharging)
```
mardy: This is equivalent to

    var isCharging = (batteryIconName.indexOf("charging") >= 0) || (deviceState != "discharging)

qml/Panel/Indicators/IndicatorsLight.qml
```diff
@@ -22,12 +23,126 @@ import Powerd 0.1
...
+            isCharging = deviceState != "discharging"
+
+        if(hasMessages) {
```
mardy: The part of the function starting with this line would be better coded as a few `State` elements: it would be easier to read and verify for correctness. (you can read about it [here](https://doc.qt.io/qt-5/qtquick-statesanimations-states.html))

mardy (dismissed):

qml/Panel/Indicators/IndicatorsLight.qml
wdehoog: The states might be easier to read but the if statements needed to determine the actual state will remain the same.

It looks like when using State the order of the property value changes is not specified. Due to the current implementation of the Lights plugin the led must be turned _off_ and _on_ after changing the color and pulse rates.

I don't mind using State but do you know of a way to get the led off/on triggered at the end?

qml/Panel/Indicators/IndicatorsLight.qml
```diff
@@ -91,55 +87,79 @@ QtObject {
...
+        var isCharging = batteryIconName.indexOf("charging") >= 0
+                         || deviceState != "discharging"
+
+        var isFull = isCharging
+                     && (batteryIconName.indexOf("full-charged") >= 0
+                         || deviceState == "fully-charged"
+                         || batteryLevel >= 100)
+
+        if (isCharging) {
+            if (isFull)
```
mardy: Due to the lines above `isFull` can only be true if also `isCharging` is true.
```
if (isFull) {
    indicatorState = "BATTERY_FULL"
} else if (isCharging) {
    indicatorState = "BATTERY_CHARGING"
} else if (batteryIconName.indexOf("caution") >= 0
           || batteryIconName.indexOf("empty") >= 0) {
    indicatorState = "BATTERY_LOW"
} else {
    indicatorState = "INDICATOR_OFF"
}

Comment (wdehoog) [reply qml/Panel/Indicators/IndicatorsLight.qml]: That is intentional. I think the battery full indication must be turned off when the charger is disconnected.

Comment (mardy) [reply qml/Panel/Indicators/IndicatorsLight.qml]: Yes, and I agree, but I'm saying that then the code can be simplified in the way I proposed.

**Review by UniversalSuperBox — CHANGES_REQUESTED** [review]
It appears that this has broken a test in Unity8...

```
FAIL!  : qmltestrunner::IndicatorsLight::test_LightsStatus(Powerd.Off with No Message) Light does not match expected value
   Actual   (): 1
   Expected (): 0
   Loc: [/build/varlibjenkinsworkspaceubportsunityPRunityubportsgbpabfdsc/tests/qmltests/Panel/Indicators/tst_IndicatorsLight.qml(129)]
```

This does not appear to be a transient error, I believe some behavior has been changed in a way you don't expect.

**Comment by wdehoog** [comment]
Sorry I give up. This testing framework is even more complicated then unity itself and the things it tests are trivial. I see no use in crafting complex dbus messages and plugins just to automatically 'test' a led turning white or green.

**Comment by wdehoog** [comment]
reopened for ci to generate the error log

File: qml/Panel/Indicators/IndicatorsLight.qml [hunk 5]
```diff
@@ -22,12 +23,140 @@ import Powerd 0.1
 import Lights 0.1
 import QMenuModel 0.1 as QMenuModel
 import Unity.Indicators 0.1 as Indicators
+import Wizard 0.1
```
mariogrip, line 26: Why does it depend on Wizard?

mariogrip: Oh i see for Powerd. We *should* move this to a common place at some point.

mariogrip (approved):
LGTM :+1:

mariogrip:
Awesome work and thank you for your contribution! :)

qml/Panel/Indicators/IndicatorsLight.qml
UniversalSuperBox: I don't think the import of Wizard is needed any more. Whatever it was imported for doesn't seem to be a problem any more. I removed the import locally and it continues to function.

UniversalSuperBox (changes requested):
Looks like one thing that could be improved.

Also, I would appreciate some new tests to ensure that this is able to return the correct states for each charging mode. However, I don't want to put that all on you. Maybe we could get together some time to work through that together?

wdehoog:
@UniversalSuperBox  Are you sure about the Wizard import? I added it due to 
```
property string batteryIconName: Status.batteryIcon
```
Status comes from the Wizard plugin. 

This weekend I will look into it and try to come up with new tests.

UniversalSuperBox:
Oh, alright. Just ping me if you need anything, you know where to find me.

UniversalSuperBox (approved):
Very nice. Excellent work. Thanks for dealing with us and our silly integration tests through this process. :)

myii:
@wdehoog Thank you for providing this.  It's reached the `devel` channel now and it's an excellent feature.  I've tested all of the indicators except for full charge (will try that next) and all working as expected.

myii:
Full battery indicator also working (solid green).  One thing is that it shows while the battery indicator in the notification bar is still green (100%) rather than the full charge white (100%).  Shouldn't it wait for that event before being displayed?

wdehoog:
@myii I noticed a really long delay (45min) between battery level reaching 100% and the signal full (icon name full-charged) being received. I decided to use the 100% level so that is why there is a difference.

giiba:
I like this feature and it seems to be working fine on my OPO. I have a suggestion/request though:

I think it would be good for the indicator light to be able to display the difference between reading 100% and the icon changing from green to white (my understanding is that reading 100% is when the charging circuit switches to a low power slow charge to top the battery fully up) as this indicates a true full charge. Also I dislike the light just staying on 😉.

I'd suggest the behaviour that, at 100% the light turns fully green, and that at 'signal full' the light turns off. This way the light indicates the charging status completely and isn't just always on when plugged in.

Danfro:
E5 RC:
charging led is white
notification pending while charging led is blinking green
fully charged led is steady green

Seems to work as expected.

cibersheep:
N5 RC:
- Low battery: to be tested
- charging led: fixed white
- Notification (charged, charging and not charging): slow blinking green
- Full charge: fixed green

M10HD:
- Low battery: led is off
- charging led: fixed white
- Notification (charged, charging and not charging): slow blinking green
- Full charge:  to be tested

PR #228, opened 2019-12-01
[Stage] Remove focus from app when in spread mode

### LINKED ISSUE #1236: Keyboard is detached from app when opening app stack
- Device: E5
- Channel: edge
- Build: 2019-09-25

### Steps to reproduce

### Expected behavior

### Actual behavior
Reproducible: always

screenshot20190925_201647608

mariogrip:
If we enter spread, we should remove focus from any app

Focus should not be given back here as that is handled by spread when
selecting app.

This fixes: https://github.com/ubports/ubuntu-touch/issues/1236
Needs: https://github.com/ubports/unity8/pull/229

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -744,7 +744,12 @@ void TopLevelWindowModel::closeAllWindows()
     }
 }
 
-void TopLevelWindowModel::rootFocus(bool focus)
+bool TopLevelWindowModel::rootFocus()
+{
+    return m_nullWindow->focused();
+}
+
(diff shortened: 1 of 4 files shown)


NeoTheThird:
3i3rld

UniversalSuperBox:
If you class 49 tests as "Some," sure.

UniversalSuperBox:
Good thing those tests are there, too. There seem to be some problems with this PR as it stands now. I think they stem from:

```
file:///home/dalton/unity8/qml/Stage/Stage.qml:499:9: QML State: Binding loop detected for property "when"
```

UniversalSuperBox:
Closing in favor of https://github.com/ubports/unity8/pull/233

PR #229, opened 2019-12-02
[TLWM] Change rootFocus to a property

mariogrip:
This changes rootFocus to a qt property instead of a callable function

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.h
+++ b/plugins/WindowManager/TopLevelWindowModel.h
@@ -88,6 +88,24 @@ class WINDOWMANAGERQML_EXPORT TopLevelWindowModel : public QAbstractListModel
      */
     Q_PROPERTY(int nextId READ nextId NOTIFY nextIdChanged)
 
+
+   /**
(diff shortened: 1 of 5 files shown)


plugins/WindowManager/TopLevelWindowModel.h
```diff
@@ -196,6 +197,8 @@ class WINDOWMANAGERQML_EXPORT TopLevelWindowModel : public QAbstractListModel
 
     void closedAllWindows();
 
+    void rootFocusChanged();
```
UniversalSuperBox, line 216: This signal is never emitted

UniversalSuperBox (changes requested):
Whilte Qt *is* magic, I think you missed a step here.

plugins/WindowManager/TopLevelWindowModel.h
mariogrip: whops, will add that :D

mariogrip (commented):
LGTM :+1:

UniversalSuperBox (approved):

PR #230, opened 2019-12-03
WIP: Top Panel Color Tweak when Dialer Is Not Visible

cibersheep:
When Dialer is not visible, top panel colored as a positive action and should be focus area colored (activity)

imatge

### DIFF
--- a/qml/Components/Showable.qml
+++ b/qml/Components/Showable.qml
@@ -74,6 +74,9 @@ Item {
     }
 
     function __reallyShow() {
+        if (showAnimation != undefined && showAnimation.running)
+            return;
+
         if (!available) {
             __skipShowAnimation = false;
             return false;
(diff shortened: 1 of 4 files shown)


UniversalSuperBox:
The contrast ratio of white-on-activity is worse than white-on-positive, as shown here:

https://contrast-ratio.com/#white-on-%233EB34F
https://contrast-ratio.com/#white-on-%2319B6EE

Minimum contrast ratio is defined at 4.5 by the [WCAG accessibility standard on contrast ratio](https://www.w3.org/TR/WCAG/#contrast-minimum). White-on-green is at 2.7 while white-on-blue is at 2.34. Do we still want to continue?

cibersheep:
Mmm... true. 
I  mean, I still believe should be activity color but I see that is getting worse for reading this blue, with white text.

Could be a good moment to start moving to the color shades Stefano was proposing some time ago

PR #232, opened 2019-12-10
Fix Notification animations

krille-chan:
Using NumberAnimations should fix this. This transitions are from the FluffyChat messages list and are working flawless.


krille-chan:
I dont understand why the pipeline fails :-( Wrong indentations?

UniversalSuperBox:
No, if you search in the log output, the error is a segmentation fault when testing the notification renderer. Something appears to be deleted then reused again, or something similar.

krille-chan:
I have experienced some small glitchy animation bugs when there are around 10 notifications at the same time. Need to rework this ...

PR #233, opened 2019-12-11
[Stage] Remove focus from app when in spread mode

### LINKED ISSUE #1236: Keyboard is detached from app when opening app stack

UniversalSuperBox:
If we enter spread, we should remove focus from any app

Focus should not be given back here as that is handled by spread when
selecting app.

This fixes: ubports/ubuntu-touch#1236

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -497,6 +497,11 @@ FocusScope {
             PropertyChanges { target: noAppsRunningHint; visible: (root.topLevelSurfaceList.count < 1) }
             PropertyChanges { target: blurLayer; visible: true; blurRadius: 32; brightness: .65; opacity: 1 }
             PropertyChanges { target: wallpaper; visible: false }
+
+            // If we enter spread, remove focus from any app
+            // Focus should not be given back here as that
+            // is handled by spread when selecting app.
+            PropertyChanges { target: topLevelSurfaceList; rootFocus: false }
         },
         State {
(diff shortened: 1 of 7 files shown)


UniversalSuperBox:
Currently in draft status because lots of tests fail. They do locally, it's not worth even letting CI run right now.

qml/Stage/Stage.qml
```diff
@@ -497,6 +497,11 @@ FocusScope {
             PropertyChanges { target: noAppsRunningHint; visible: (root.topLevelSurfaceList.count < 1) }
             PropertyChanges { target: blurLayer; visible: true; blurRadius: 32; brightness: .65; opacity: 1 }
             PropertyChanges { target: wallpaper; visible: false }
+
+            // If we enter spread, remove focus from any app
+            // Focus should not be given back here as that
+            // is handled by spread when selecting app.
+            PropertyChanges { target: topLevelSurfaceList; rootFocus: false }
```
UniversalSuperBox: It appears that this change is already done with the stage's interactive property: https://github.com/ubports/unity8/blob/5095f7e4b4a762b47aa5c23c6c51398f77977290/qml/Stage/Stage.qml#L103

UniversalSuperBox:
Closing in favor of bringing all these changes into https://github.com/ubports/unity8/pull/235, trying to maintain two clean branches is proving too much for my mind.

PR #235, opened 2019-12-12
TopLevelWindowModel: Don't refocus closed Window

### LINKED ISSUE #234: Unity8 crashes if focused app is closed when rootFocus is given back
TopLevelWindowModel's rootFocus property tracks the app that was focused before rootFocus was taken away. This allows it to give the app focus back when rootFocus is set.

However, if the app that was previously focused is closed before rootFocus is given back, and no activation is pending, Unity8 crashes with a segmentation fault.

This can be reproduced by adding the following test to `tst_PhoneStage.qml`:

```
function test_closeFocusedAppWithSpreadOpen()
{
    var dashApp = ApplicationManager.findApplication("unity8-dash");
    var webbrowserSurfaceId = topLevelSurfaceList.nextId;
    var webbrowserApp  = ApplicationManager.startApplication("morph-browser");
    waitUntilAppSurfaceShowsUp(webbrowserSurfaceId);

    topLevelSurfaceList.rootFocus = false;

    performEdgeSwipeToShowAppSpread();

    swipeSurfaceUpwards(webbrowserSurfaceId);
    tryCompareFunction(function() { return topLevelSurfaceList.indexForId(webbrowserSurfaceId); }, -1);

    topLevelSurfaceList.rootFocus = true;
    compare(dashApp.focused, true);
}
```

Currently in `xenial` this code path is not hit.

UniversalSuperBox:
I'm not sure if we should refocus the next window in this case... but I don't think there's anything else we can do. The client told TLWM that they want rootFocus to be set, so we have to focus something other than the null window... otherwise rootFocus won't change.

Prior to this, clearing rootFocus, closing the previously focused Window, then setting rootFocus would cause a use-after-delete, resulting in a segfault at best or undefined behavior at worst.

Fixes https://github.com/ubports/unity8/issues/234

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -38,7 +38,8 @@ Q_LOGGING_CATEGORY(TOPLEVELWINDOWMODEL, "toplevelwindowmodel", QtInfoMsg)
 namespace unityapi = unity::shell::application;
 
 TopLevelWindowModel::TopLevelWindowModel()
-    : m_nullWindow(createWindow(nullptr))
+    : m_nullWindow(createWindow(nullptr)),
+      m_surfaceManagerBusy(false)
 {
     connect(m_nullWindow, &Window::focusedChanged, this, [this] {
         Q_EMIT rootFocusChanged();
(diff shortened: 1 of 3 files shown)


UniversalSuperBox:
Closing to make room for *yet another* PR trying to fix the same problem, sorry for the spam, watchers.

PR #239, opened 2019-12-13
Remove focus from the topmost app when in spread

### LINKED ISSUE #234: Unity8 crashes if focused app is closed when rootFocus is given back

### LINKED ISSUE #1236: Keyboard is detached from app when opening app stack

UniversalSuperBox:
By changing the semantics of the Stage's 'interactive' property, we're able to remove focus from the topmost app when the Spread is open and give it back at an appropriate time.

This exposed #234, which had to be fixed first.

Fixes https://github.com/ubports/ubuntu-touch/issues/1236

Fixes https://github.com/ubports/unity8/issues/234

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -36,7 +36,6 @@ FocusScope {
     property bool altTabPressed
     property url background
     property int dragAreaWidth
-    property bool interactive
     property real nativeHeight
     property real nativeWidth
     property QtObject orientations
@@ -48,6 +47,11 @@ FocusScope {
     property rect inputMethodRect
     property real rightEdgePushProgress: 0
     property Item availableDesktopArea
+
+    // Whether outside forces say that the Stage may have focus
+    property bool allowInteractivity
+
+    readonly property bool interactive: (state === "staged" || state === "stagedWithSideStage" || state === "windowed") && allowInteractivity
 
     // Configuration
     property string mode: "staged"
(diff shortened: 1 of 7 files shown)


qml/Stage/Stage.qml
```diff
@@ -89,18 +93,17 @@ FocusScope {
                 Qt.InvertedLandscapeOrientation;
     }
 
+    Binding {
+        target: topLevelSurfaceList
+        property: "rootFocus"
+        value: interactive
+    }
+
```
mariogrip, line 101: Are you sure this wont cause a condition where rootFocus is set when stage does not have focus? Since stage will need to have focus before activation of windows, this includes the null window.

mariogrip (approved):
other then my little comment, LGTM :+1:

qml/Stage/Stage.qml
UniversalSuperBox: interactive is cleared whenever the Stage is not in staged, stagedWithSideStage, or windowed state AND allowInteractivity is true: https://github.com/ubports/unity8/pull/239/files#diff-575a1ba4a622df698d6c1dd36cf9f5f1R54

allowInteractivity is set only when the greeter is not active, indicators are closed, notifications are not in 'modal' mode, and the Launcher isn't taking focus: https://github.com/ubports/unity8/pull/239/files#diff-a54a1698b32808661587b74b7e8ac356R326

If interactive is cleared, rootFocus is false.

What case do you see rootFocus being set without the Stage having focus?

Edit: Also, if rootFocus is set, the various activations between surfaces and windows will ensure that the Stage has focus before they all finish. :P

PR #244, opened 2019-12-16
[Stage] handle "Application" level focus request

### LINKED ISSUE #1242: Killed apps cannot be refocused or bring to the foreground
- Device: Nexus 5
- Channel: Edge
- Build: 2019-10-03/3

### Steps to reproduce
1. Open lots of apps until an app gets killed. You can validate from the launcher if an unpinned app has no dots in both left and right side of the icon.
2. Tap on that app with no dots.

### Expected behavior
App gets focused and reopened.

### Actual behavior
Nothing happens and the app doesn't reopen.

peat-psuwit:
Application emits focusRequested when it has no surface (i.e. their
processes died). Unity 8 used to handle this case. However, commit
f1dcc66d18e1218ebfa46f205681048a4e71989f (Let the model deal with some
window management decisions) removed the code.

This commit makes the stage listens to each application's focusRequested
signal. When received, it'll find the topmost window for the application
and activate it, after which the app will be requested to be running.

Fixes: https://github.com/ubports/ubuntu-touch/issues/1242

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -485,6 +485,26 @@ FocusScope {
                             ? (!model.application.isTouchApp || isExemptFromLifecycle(model.application.appId))
                             : false
             }
+
+            property var focusRequestedConnection: Connections {
+                target: model.application
+
+                onFocusRequested: {
+                    // Application emits focusRequested when it has no surface (i.e. their processes died).
+                    // Find the topmost window for this application and activate it, after which the app
+                    // will be requested to be running.
+
+                    for (var i = 0; i < appRepeater.count; i++) {
+                        var appDelegate = appRepeater.itemAt(i);
+                        if (appDelegate.application.appId === model.application.appId) {
+                            appDelegate.activate();
+                            return;
+                        }
+                    }
+
+                    console.warn("Application requested te be focused but no window for it. What should we do?");
+                }
+            }
         }
     }
 


mariogrip:
I wonder if it would be better to move this logic to the cpp side https://github.com/ubports/unity8/blob/xenial/plugins/WindowManager/TopLevelWindowModel.cpp#L226

peat-psuwit:
Maybe, maybe not, but probably not in `TopLevelWindowModel`. TLWM seems to concern about windows & surfaces. However, this is in the "Application" layer, which is when the user presses the app's icon in the launcher.

UniversalSuperBox:
One of those cases where the Stage and Launcher's window model being the same would be helpful...

I'm not sure of any better way to do this in C++ without adding to our coupling problem, since the TLWM knows about the suspended Windows, and the Launcher's model does (/should) too, but the Application that the Launcher signals should not.

mariogrip:
TLWM alredy handle the application level, and a focus request should be handled by the WM imo

I would do something like this
```
diff --git a/plugins/WindowManager/TopLevelWindowModel.cpp b/plugins/WindowManager/TopLevelWindowModel.cpp
index f999edf4ad..3e6404b5c6 100644
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -125,6 +125,14 @@ void TopLevelWindowModel::addApplication(unityapi::ApplicationInfoInterface *app
     if (application->state() != unityapi::ApplicationInfoInterface::Stopped && application->surfaceList()->count() == 0) {
         prependPlaceholder(application);
     }
+
+    connect(application, &unityapi::ApplicationInfoInterface::focusRequested, this, [this, application]() {
+        for (auto windowModel : m_windowModel) {
+            if (windowModel.application == application && windowModel.window) {
+                windowModel.window->activate();
+            }
+        }
+    });
 }
 
 void TopLevelWindowModel::removeApplication(unityapi::ApplicationInfoInterface *application)
@@ -141,6 +149,8 @@ void TopLevelWindowModel::removeApplication(unityapi::ApplicationInfoInterface *
             ++i;
         }
     }
+
+    disconnect(application, 0, this, 0);
 }
 
 void TopLevelWindowModel::prependPlaceholder(unityapi::ApplicationInfoInterface *application)
```

mariogrip:
in the c++ way, we could also handle case where window is missing, like adding a placeholder if the application is restarting etc

peat-psuwit:
Hmm......

TLWM seems to care about applications being added or removed (so that it can create a placeholder window) but not about what happens inside the application. Meanwhile, Stage seems to bind into applications' lifecycle properties.

Also, most of the time Stage will be the one activating windows, including when the focus is requested at the QtMir's mirsurface level. TLWM will activate a window when the list of apps has changed (or when changing rootFocus, but that's a special case).

https://github.com/ubports/unity8/blob/efd3bd14c91f018286c679a06aedad9a9affae63/qml/Stage/Stage.qml#L984-L994

NeoTheThird:
what's the plan here?

mariogrip:
> Hmm......
> 
> TLWM seems to care about applications being added or removed (so that it can create a placeholder window) but not about what happens inside the application. Meanwhile, Stage seems to bind into applications' lifecycle properties.
> 
> Also, most of the time Stage will be the one activating windows, including when the focus is requested at the QtMir's mirsurface level. TLWM will activate a window when the list of apps has changed (or when changing rootFocus, but that's a special case).
> 
> https://github.com/ubports/unity8/blob/efd3bd14c91f018286c679a06aedad9a9affae63/qml/Stage/Stage.qml#L984-L994

I would like to abstract as much of the window management handling to c++ as it's much faster, imo qml should not do any WM job other then telling the WM what to do. but this might be fine for now as this will get a heavy change once proper multi screen lands.

UniversalSuperBox (approved):
Both proposed implementations are functionally equivalent. The speedup will be a few calls in the stack at best. I think we can agree that *neither* file is where we really expect application+window management changes to go, so neither has a great impact on maintainability.

And, well, guess all this might change when we no longer use MirAL, so let's rejoice about that? Hhhheh?

PR #245, opened 2019-12-16
[DO NOT MERGE] enable debug build type for qml debugging

peat-psuwit:


dobey:
I think this is not necessary. dpkg builds as debug by default, and then strips the binaries, so it can create dbgsym packages with debug symbols.

peat-psuwit:
This is not for C/C++ debugging symbol. This is for enabling QML debugging mode.

https://github.com/ubports/unity8/blob/efd3bd14c91f018286c679a06aedad9a9affae63/CMakeLists.txt#L20-L24

See also https://doc.qt.io/qt-5/qtquick-debugging.html#qml-debugging-infrastructure.

However, I always get a crash in the Qt core whenever I enabled it.

PR #253, opened 2019-12-20
[Stage] Make sure we send screenOffsets to mir

### LINKED ISSUE #1095: App scaling is wrong on external displays
- Device: Nexus 5
 - Channel: Edge
 - Build: 16.04 (2019-04-10/2)

### Steps to reproduce
1. Connect the device to an external display i.e. Nexus 5 via slimport adapter
2. Open apps.

### Expected behavior
The apps should scale proportionate to the Unity 8 shell.

### Actual behavior
The apps are scaled bigger than the scaling of the Unity 8 shell.

mariogrip:
This makes sure we send screenOffsets to mir, if not mir belives we are
on the screen at pos 0,0 regardless of what screen we are on. This will
also make per-screen-scaling work correctly! :D

This fixes: https://github.com/ubports/ubuntu-touch/issues/1095

### DIFF
--- a/qml/Stage/Stage.qml
+++ b/qml/Stage/Stage.qml
@@ -15,6 +15,7 @@
  */
 
 import QtQuick 2.4
+import QtQuick.Window 2.2
 import Ubuntu.Components 1.3
 import Unity.Application 0.1
 import "../Components/PanelState"
@@ -767,11 +768,19 @@ FocusScope {
                 }
                 z: normalZ
 
+                // Set these as propertyes as they wont update otherwise
+                property real screenOffsetX: Screen.virtualX
+                property real screenOffsetY: Screen.virtualY
+
                 // Normally we want x/y where the surface thinks it is. Width/height of our delegate will
                 // match what the actual surface size is.
                 // Don't write to those, they will be set by states
-                x: model.window.position.x - clientAreaItem.x
-                y: model.window.position.y - clientAreaItem.y
+                // --
+                // Here we will also need to remove the screen offset from miral's results
+                // as unity8 x,y will be relative to the current screen only
+                // FIXME: when proper multiscreen lands
+                x: model.window.position.x - clientAreaItem.x - screenOffsetX
+                y: model.window.position.y - clientAreaItem.y - screenOffsetY
                 width: decoratedWindow.implicitWidth
                 height: decoratedWindow.implicitHeight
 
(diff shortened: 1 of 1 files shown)


qml/Stage/Stage.qml
```diff
@@ -767,11 +768,15 @@ FocusScope {
...
+                property real screenOffsetX: Screen.virtualX
+                property real screenOffsetY: Screen.virtualY
+
                 // Normally we want x/y where the surface thinks it is. Width/height of our delegate will
                 // match what the actual surface size is.
                 // Don't write to those, they will be set by states
-                x: model.window.position.x - clientAreaItem.x
-                y: model.window.position.y - clientAreaItem.y
+                x: model.window.position.x - clientAreaItem.x - screenOffsetX
+                y: model.window.position.y - clientAreaItem.y - screenOffsetY
```
peat-psuwit: Isn't this X/Y relative to the parent item i.e. the stage i.e. current screen?

qml/Stage/Stage.qml
```diff
@@ -784,10 +789,15 @@ FocusScope {
                 Binding {
                     target: model.window; property: "requestedPosition"
                     // miral doesn't know about our window decorations. So we have to deduct them
-                    value: Qt.point(appDelegate.requestedX + appDelegate.clientAreaItem.x,
-                                    appDelegate.requestedY + appDelegate.clientAreaItem.y)
+                    value: Qt.point(appDelegate.requestedX + appDelegate.clientAreaItem.x + screenOffsetX,
+                                    appDelegate.requestedY + appDelegate.clientAreaItem.y + screenOffsetY)
```
peat-psuwit, line 801: You may want to comment here that "the position is relative to Mir's global coordinate i.e. screen offset is required".

peat-psuwit (changes requested):
I have a doubt and a suggestion.

qml/Stage/Stage.qml
mariogrip: So unity8's and mir positions are different, unity8 uses the position of just the current screen (only once screen can have a shell currently)

unity8 x,y will be relative to the current screen only, so we need to remove the offset that comes from model.window.position as they will include the offset

An example, where we have two screens, screen 1 at 0,0 and screen 2 at 1200,0, and we place a application at the second screen at 1210,0

Unity8 will then place this as x,y 10,0 on screen 2, while mir will place this at 1210,0. So in simple term, unity8 x,y have no idea what screen its on as unity8's shell currently can only live on one screen anyway

This will be fixed by the really big pr: https://github.com/ubports/unity8/pull/164, but that wont be merged before OTA-12 has landed, so this is quick hack for the current version to have correct scaling in convergence mode.

qml/Stage/Stage.qml
mariogrip: Will add some comments

qml/Stage/Stage.qml
peat-psuwit: Just noticed, it's "-", not "+". Side effect of working too late. :D

peat-psuwit (approved):
Sorry, got occupied by preparing things for 36C3.

It's LGTM now. However, I think you'll need to retry building this commit.

mariogrip:
> Sorry, got occupied by preparing things for 36C3.
> 
> It's LGTM now. However, I think you'll need to retry building this commit.

Ah! I have forgotten you where going to 36c3 :D but no worries :) I hope you will enjoy 36c3 :)

Yeah i'll try a rebuild, the tests seems (bit too) flaky but thats unrelated to this pr.

PR #254, opened 2019-12-24
Fix Unity8 tests again (again?)

UniversalSuperBox:
Merry Christmas, everyone! For your present, I fixed the build. Ignore the fact that I broke the build. *whistles nonchalantly*

Just some more race conditions and interesting behavior. In particular, the TopLevelWindowModel is capable of issuing ~~1 million~~ (edit: now 2 billion) unique window IDs but instead sometimes issues the same ID for multiple windows. Fun! With some atomicity added, this case should happen much less often now.

Then we continue on, fix a couple of longer-standing problems in the test suite and do some refactoring to help reduce some race conditions leading to tests that fail *sometimes*. Finish off by quieting down a ton of warnings from fake indicators and we finish with a slightly more sane test battery.

Please either **merge with a merge commit** or **rebase and merge**, the history is meaningful.


UniversalSuperBox:
*On the 25th day of Christmas, Jenkins CI said to me,*
`This commit cannot be built`

UniversalSuperBox:
Finally, all passing. Ready for review.

plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -38,7 +39,8 @@ Q_LOGGING_CATEGORY(TOPLEVELWINDOWMODEL, "toplevelwindowmodel", QtInfoMsg)
 namespace unityapi = unity::shell::application;
 
 TopLevelWindowModel::TopLevelWindowModel()
-    : m_nullWindow(createWindow(nullptr)),
+    : m_nullWindow(createNullWindow()),
+      m_maxWindowId(1000000),
```
dobey: I presume this is changed to a non-const int to be able to initialize or set to a very small value for tests. However, the old comment below is I think wrong and should be removed, and we should use `std::numeric_limits<int>::max()` here, as Qt seems to use a 32-bit int for the QML int.

dobey (changes requested):

plugins/WindowManager/TopLevelWindowModel.cpp
UniversalSuperBox: It is to allow for messing with it in tests, yes, and now I see the QML documentation says "The possible int values range from around -2000000000 to around 2000000000", the range of a 32-bit signed integer. I'll make that change.

UniversalSuperBox: Got it, and tests passed locally. Now we wait for `-j1` tests to do their work...

mariogrip (approved):
LGTM :+1:

dobey (approved):

PR #257, opened 2019-12-30
Remove logic to wrap around from maxWindowId to 1

UniversalSuperBox:
Now that we can handle up to around 2 billion window IDs at a time, trying to wrap around from the maximum back to 1 is a little excessive. You'd have to open a window every second for 63 years in order to overflow the max window ID.

Also the logic is a bit flawed, it never checked if the ID it was about to issue was already in use *after* the time we wrapped around at 1. So, if you had ID 3 open but ID 4 in use, it would happily issue ID 4 twice.


dobey (approved):

mariogrip (approved):
LGTM :+1:

PR #263, opened 2020-01-11
Fix theme in ShellDialog

cibersheep:
Make ShellDialog to use Ambiance instead of a theme hack

This change is related to https://github.com/ubports/ubuntu-ui-toolkit/pull/64

### DIFF
--- a/qml/Components/ShellDialog.qml
+++ b/qml/Components/ShellDialog.qml
@@ -15,9 +15,7 @@
  */
 
 import QtQuick 2.4
-
 import Ubuntu.Components 1.3
-import Ubuntu.Components.Themes 1.3
 import Ubuntu.Components.Popups 1.3
 import Utils 0.1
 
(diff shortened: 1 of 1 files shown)


cibersheep:
With the changes in uitk, and this MR, Power off dialog will look like (both Ambiance and SuruDark):

image

mardy (approved):
Looks good to me. Just wondering though, why do we want to enforce a theme? How would the power button dialog look like in the dark theme, without this theme override?

cibersheep:
> Looks good to me. Just wondering though, why do we want to enforce a theme? How would the power button dialog look like in the dark theme, without this theme override?

Thanks.

I just learn recently that the Shell (as its child elements) is SuruDark. Like that panel, indicators, etc are SuruDark. Then the system «forces» to a default theme (which by design was Ambiance) and the uitk writes to that into `./config/unity8/theme.ini`. We take advantage of that ini file and modify it to make the system SuruDark or Ambiance.

My idea was to make ShellDialog and Notification bubbles to be themed as the user choses but I couldn't find the way to read the ini file, as reading `theme.name` give the parent theme (Shell which is SuruDark). So, I left the Ambiance definition here for now and I'll check if setting the default theme in the gsettings schema is the right way of telling the system what's the default theme user choses.

Sorry for the loooong answer. The sort answer is: If we don't set Ambiance here, ShellDialog would be always themed as SuruDark.

PS: In my MR I changed the hack (not needed anymore) and replaced it by the «correct» way of subtheming just as a transition.

Any thoughts, suggestions or fixes are welcome of course.

UniversalSuperBox (approved):
When installed with https://github.com/ubports/ubuntu-ui-toolkit/pull/64, this fixes the problem I mentioned in my review there.

PR #267, opened 2020-01-24
[TLWM] always activate the newly-prepended window

peat-psuwit:
After prepending a window, the window should be activated to make sure
MirAL's focused window match the top window. There's no need to check if
there's a surface, as Window::activate() will check that by itself.

Test case:
1. Connect a keyboard.
2. Open Morph Brower. Then, create a new window.
3. Press Tab a few times. The focus inside the new window should
   eventually go to the address bar.

To properly regress-test this, the test mock is also updated to mimick qtmir more closely. This requires updates on a few tests. See the detail in the commit message.

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -195,9 +195,8 @@ void TopLevelWindowModel::prependSurfaceHelper(unityapi::MirSurfaceInterface *su
 
     prependWindow(window, application);
 
-    if (!surface) {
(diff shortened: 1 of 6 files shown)


mariogrip (approved):
LGTM :+1:

peat-psuwit:
Tested on Halium based FP2.

peat-psuwit:
Ah, shit. This PR breaks emergency calls from the greeter. Do not merge yet.

**Edit:** Actually, no. It's another thing. Go ahead and merge it as you see fit.

PR #273, opened 2020-02-12
WIP: allow greeter rotation

mateosalta:
quick hack for rotation on lock screen

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -76,7 +76,7 @@ StyledItem {
 
     readonly property bool orientationChangesEnabled: panel.indicators.fullyClosed
             && stage.orientationChangesEnabled
-            && (!greeter || !greeter.animating)
+            && (!greeter.animating)
 
     readonly property bool showingGreeter: greeter && greeter.shown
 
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
It should be noted that this breaks the NarrowView greeter -- that is, the phone one, and *really* breaks the WideView used on tablets and desktops.

Narrow view with wide screen

Wide view with narrow screen


Adding rotation to the lock screen will require a small redesign of the way the Greeter is loaded and laid out. In fact, we might need to combine the narrow and wide views into a single file, then lay them out with clever states and transitions. This will require a lot of care as well, since the Greeter is a secured area of the system. We need to make sure we're protecting the security of the device and the privacy of its user.

UniversalSuperBox (changes requested):
Blocking review for above comment

mateosalta:
I agree, we should design it to be one convergent layout. very work in progress so far :) Hope this will really help others work towards this, as previously it was theorized that the greeter couldn't rotate or was constrained by a static file. since this is bypassed we can make progress on designing a new convergent greeter ;)

Also I'm thinking we should respect the primary orientation perhaps for a half second - think of picking up a device, or glancing at a phone at odd angle. let it wake up primary then smoothly transition to current orientation

Fuseteam:
This is superceded by #402

PR #274, opened 2020-02-12
Lock Launcher on screen when no apps are running

### LINKED ISSUE #133: Unpin of shortcut while no apps running hides launcher
While the launcher is being shown as no apps are running, if one un-pins an app from the launcher, it will slide off screen. It should remain on-screen in this case.

UniversalSuperBox:
This commit changes the behavior of the auto-hiding Launcher, making it show whenever there are no apps running and the full-screen indicator panel is closed.

This fixes what was mostly a design issue, where users could dismiss the Launcher by tapping on the Background, losing the Launcher entirely.

Additionally, fixes https://github.com/ubports/unity8/issues/133

### DIFF
--- a/tests/qmltests/tst_Shell.qml
+++ b/tests/qmltests/tst_Shell.qml
@@ -577,7 +577,7 @@ Rectangle {
         function loadShell(formFactor, apps) {
 
             if (apps === undefined) {
-                apps = 0
+                apps = 0;
             }
 
             shellLoader.state = formFactor;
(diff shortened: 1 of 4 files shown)


tests/qmltests/tst_Shell.qml
```diff
@@ -3246,5 +3246,61 @@ Rectangle {
...
+            // dismisses in a failure case.
+            tap(shell);
+            wait(100);
+            tryCompare(launcher, "state", "visible");
+            tryCompare(launcher, "x", 0);
+
+            // The Launcher should dismiss at phone size, but not other sizes,
+            // when the indicators open
+            showIndicators();
+            if (data.formFactor === "phone") {
```
dobey: I think it would be better to have the state value as part of the test data, rather than introducing logic (if statements) into the tests.

dobey (changes requested):

UniversalSuperBox:
I didn't expect to force-push, but I realized that I had messed up one of the test cases.

With all of that data, I expanded the objects. Otherwise the lines were way too long.

tests/qmltests/tst_Shell.qml
```diff
@@ -3246,5 +3246,93 @@ Rectangle {
...
+                stateWithOpenApps: ""
+            },
+            {
+                tag: "tablet with auto-hide at small formFactor",
+                formFactor: "phone",
+                shell: "tablet",
+                launcherLocked: false,
+                stateWithOpenIndicators: "",
+                stateWithOpenApps: ""
+            },
```
dobey: Missing closing bracket for the list of objects.

dobey (changes requested):

tests/qmltests/tst_Shell.qml
UniversalSuperBox: Queue "how did this even work for me before" response.

dobey (approved):

PR #276, opened 2020-02-18
Prevent IndicatorItem removal while expanding

### LINKED ISSUE #132: Confusing Animation When Pulling down Indicator

peat-psuwit:
PanelBar uses some tricks to maintain the initial item's position while
all items are expanding. For that to work, the initial item (and maybe
every item, but I'm not sure) must not be removed by the ListView
because PanelBar relies on the item's position for calculation.

Ideally, because PanelItemRow expands the ListView to match its
children, no item should fall out of its view. However, upon testing,
the rightmost item can fall out of view and get destroyed under some
circumstances.

To solve this, ListView.delayRemove is bound to the transition's running
state, so that it won't be removed while it's expanding (or retracting,
but that probably doesn't matter).

Fixes ubports/unity8#132.

### DIFF
--- a/qml/Panel/Indicators/IndicatorItem.qml
+++ b/qml/Panel/Indicators/IndicatorItem.qml
@@ -38,6 +38,11 @@ IndicatorDelegate {
 
     implicitWidth: mainItems.width
 
+    // Prevent ListView from removing us from the view while expanding.
+    // If we're the PanelBar's initial item, our removal will make it lose
+    // track of us and cause its positioning to go wrong.
+    ListView.delayRemove: stateTransition.running
+
     MouseArea {
         readonly property int stepUp: 1
         readonly property int stepDown: -1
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
This may additionally fix https://github.com/ubports/ubuntu-touch/issues/1364, but I have been unable to reproduce the problem.

cibersheep:
It's an absolute pleasure to use indicators with this fix. Thanks
I have tried to reproduce the https://github.com/ubports/ubuntu-touch/issues/1364 and I couldn't. I'll use it the next days

UniversalSuperBox (approved):
This makes indicators almost as pleasing to use as they were prior to the Unity8 upgrade. Thank you, thank you, thank you!

Also, the pr-merge job was canceled, it didn't fail. The correct build is here: https://ci.ubports.com/blue/organizations/jenkins/ubports%2Funity8/detail/PR-276/6/pipeline

PR #278, opened 2020-02-18
Fix Launcher dismissing keyboard

### LINKED ISSUE #1370: reopen #1245 lower part of launcher not accessible when OSK is raised
This issue is to reopen #1245 because it is still valid. This issue was not fixed with [ubports/unity8#206](https://github.com/ubports/unity8/pull/206). While I can not reopen the original issue, I create a new one, so we don't miss this out.

Device: E5
Channel: devel
Build: 2020-02-14

Steps to reprocuce:

- open launcher
- type something in search bar to raise the OSK
- swipe the launcher towards the icons in the lower part of the launcher

Expected behaviour:

- all icons can be reached

Current behaviour:

- the lower icons are hidden behind the OSK, the launcher can not be scrolled low enough to reach them

Possible solutions:
- When the launcher is moved OR when the search bar looses the focus the OSK is dismissed
- The launcher is resized so its bottom matches top of OSK

UniversalSuperBox:
329a3859 attempted to make dragging on the Launcher dismiss the keyboard. Unfortunately, it relied on the LauncherPanel's 'dragging' property being set correctly.

This commit fixes the 'dragging' property so the behavior works correctly.

This *will* cause the keyboard to be dismissed when desktop users who have the launcher locked drag the launcher. I think this is acceptable behavior, if you disagree please comment.

Fixes https://github.com/ubports/ubuntu-touch/issues/1370

### DIFF
--- a/qml/Launcher/LauncherPanel.qml
+++ b/qml/Launcher/LauncherPanel.qml
@@ -31,13 +31,13 @@ Rectangle {
     property var model
     property bool inverted: false
     property bool privateMode: false
-    property bool dragging: false
     property bool moving: launcherListView.moving || launcherListView.flicking
     property bool preventHiding: moving || dndArea.draggedIndex >= 0 || quickList.state === "open" || dndArea.pressed
                                  || dndArea.containsMouse || dashItem.hovered
     property int highlightIndex: -2
     property bool shortcutHintsShown: false
     readonly property bool quickListOpen: quickList.state === "open"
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
The pr-merge job did not fail, it was canceled. The last job was https://ci.ubports.com/blue/organizations/jenkins/ubports%2Funity8/detail/PR-278/6/pipeline

NeoTheThird (approved):
Seems to work fine, can be merged if the pipeline succeeds.

PR #279, opened 2020-02-20
Make Tutorial start faster; block functionality

### LINKED ISSUE #746: Introductory tour is too slow
- Device: Frieza, Hammerhead
- Channel: Devel
- Build: 2018-07-04

### Steps to reproduce
Wipe and reinstall Ubuntu Touch

### Expected behavior
The introductory tour should proceed once you have configured your account. One instruction / hint should follow briskly after another.

### Actual behavior
The first instruction [swipe from the Left edge] doesn't occur for aprox 30 seconds. It could be confusing for a new user because it isn't clear what triggers it.
Subsequent instructions occur after a delay.

Is there something useful I can provide here?

UniversalSuperBox:
These changes make the Tutorial far tighter. It starts faster, demonstrates what it needs to, then gets out of the way.

This fixes the most common complaint about the Tutorial, that it started too slowly. It also puts the steps in (what I believe) is a more logical order. First Indicators, then the Launcher, then Drawer, then Spread after the user opens an app. The Launcher is disabled until its tutorial appears, likewise for the Drawer.

Fixes https://github.com/ubports/ubuntu-touch/issues/746

https://github.com/ubports/ubuntu-touch/issues/362 is a separate issue not fixed by this commit, in case anyone was wondering since I mentioned it recently.

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -580,7 +580,7 @@ StyledItem {
             panelWidth: units.gu(settings.launcherWidth)
             lockedVisible: (lockedByUser || shell.atDesktop) && lockAllowed
             topPanelHeight: panel.panelHeight
-            drawerEnabled: !greeter.active
+            drawerEnabled: !greeter.active && tutorial.launcherLongSwipeEnabled
             privateMode: greeter.active
             background: wallpaperResolver.background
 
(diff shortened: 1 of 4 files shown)


dobey (approved):
Nice.

PR #281, opened 2020-02-20
Allow user setting "autohideLauncher" to set prefrence

mateosalta:
Check if user has set autohideLauncher for locked launcher preference.

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -588,12 +588,12 @@ StyledItem {
             // the Panel is open and taking up the full width of the shell
             readonly property bool collidingWithPanel: panel && (!panel.fullyClosed && !panel.partialWidth)
 
-            // The "autohideLauncher" setting is only valid in desktop mode
+            // The "autohideLauncher" setting for desktop mode
             readonly property bool lockedByUser: (shell.usageScenario == "desktop" && !settings.autohideLauncher)
 
             // The Launcher should absolutely not be locked visible under some
-            // conditions
-            readonly property bool lockAllowed: !collidingWithPanel && !panel.fullscreenMode && !wizard.active
+            // conditions, allow user setting "autohideLauncher" to override lock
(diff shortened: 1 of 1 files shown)


mateosalta:
fix https://github.com/ubports/unity8/issues/282

UniversalSuperBox (changes requested):
This patch appears to put off the problem, not fix it. When a user wants to auto-hide the launcher on the desktop, they do not necessarily want it to always auto-hide on mobile.

It's very likely that this will fail tests. Even if it does not, the tests will need to be modified to ensure that every combination of auto-hide or not, phone, tablet, or desktop is covered. See https://github.com/ubports/unity8/blob/xenial/tests/qmltests/tst_Shell.qml#L3243 for this test.

dobey:
As @UniversalSuperBox said, this doesn't actually provide what you are thinking it does. Instead it complicates the meaning of the setting (meaning, you must set it to always, but then the launcher isn't always on). I don't think it's a reasonable compromise either, for the complaint you've put forward, as the setting should already be set to this by default, which means all you're doing is restoring the previous behavior for everyone by default.

mateosalta:
> As @UniversalSuperBox said, this doesn't actually provide what you are thinking it does. Instead it complicates the meaning of the setting (meaning, you must set it to always, but then the launcher isn't always on). I don't think it's a reasonable compromise either, for the complaint you've put forward, as the setting should already be set to this by default, which means all you're doing is restoring the previous behavior for everyone by default.

Ah, that would be a deal breaker for sure. I thought autohide was off by default...

Fuseteam:
I'm somehow two years late on this idea.... 

> It complicates the meaning of the setting (meaning, you must set it to always, but then the launcher isn't always on). 

@dobey isn't this currently the case? The default setting for the launcher is "always hide", but it is never hidden on phones and tabets(?). Wouldn't it be better to set the default setting to "always shown" and have the user set "always hide" across staged and desktop mode?

PR #283, opened 2020-02-22
Allow to suppress charging state notification

wdehoog:
Some people don't like charging state indication and some devices might only have one led blink pattern. These changes allow to disable showing them (dconf write ```/com/lomiri/ledindication/charging-state-visible false```).

See https://github.com/ubports/ubuntu-touch/issues/1377

(edited to show new settings namespace)

### DIFF
--- a/data/com.canonical.Unity8.gschema.xml
+++ b/data/com.canonical.Unity8.gschema.xml
@@ -72,4 +72,12 @@
       <description>If too many unsuccessful login attempts have been made in a row, the greeter locks you out for a while.  This unix timestamp indicates when you were locked out, so unity8 knows when to let you back in.</description>
     </key>
   </schema>
+
+  <schema path="/com/canonical/unity8/ledindication/" id="com.canonical.Unity8.LedIndication" gettext-domain="unity8">
+    <key type="b" name="charging-state-visible">
+      <default>true</default>
+      <summary>Whether the battery charging state should be shown using the notification led</summary>
+      <description>Toggle the visibility of the battery charging state in the notification led patterns</description>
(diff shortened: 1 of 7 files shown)


UniversalSuperBox:
Is your plan to add these settings somewhere in `system-settings` as well?

Since some devices only have a single-color LED, I'd recommend adding something to the device config parser in another PR. This additional metadata could tell the system if it should even try to use multi-color LEDs. Then users don't need to choose if they want their LED on all the time during charging.

data/com.canonical.Unity8.gschema.xml
```diff
@@ -72,4 +72,12 @@
       <description>If too many unsuccessful login attempts have been made in a row, the greeter locks you out for a while.  This unix timestamp indicates when you were locked out, so unity8 knows when to let you back in.</description>
     </key>
   </schema>
+
```
mariogrip, line 75: I suggest when adding new gsettings use the com.lomiri userspace to ease with transition later on.

UniversalSuperBox: Is that possible when we're in the com.canonical.Unity.gschema.xml file?

UniversalSuperBox: Alright, now there's a new com.lomiri.gschema.xml file. @wdehoog, is the change in this file needed any more?

UniversalSuperBox:
Sorry for the wait... Unfortunately, this PR now conflicts on fake_gsettings... Could you please resolve this conflict (you may want to `git pull --rebase upstream xenial`) so we can take another look?

PR #289, opened 2020-03-04
Allow indicators to take up top panel (and also fix indicators)

### LINKED ISSUE #1233: Edge:  indicator menu, touch area too restrictive
### Illustrations

- Device: Nexus 5
- Channel: Edge
- Build: today 

### Steps to reproduce

The area before the first icon of the indicator menu is not responsive.

### Expected behavior

I would expect to have indicator menu displayed without pointing precisely to the notification icon

### Actual behavior

Indicator not displayed.

### LINKED ISSUE #1364: Indicator menuitem selects different settings panel
- Device: BQ Aquaris E4.5
- Channel: edge
- Build: 2020-02-09 (JU12489)

### Steps to reproduce
1. Switch the device on
2. Use it for a while
3. Use the indicator menu
4. After a while touching a menuitem selects different settings panel  than expected.
For example touch 'Keyboard Layout', and the device opens 'Notification' instead. Select 'Time and Date' and the 'Network' gets selected.

The slip between touching point and displayed menu is increasing over time. After a while it gets unusable.
Rebooting the device resets the behaviour (kind of workaround).

### Expected behavior
The indicator menu should display exactly the content of selected menuitem.

UniversalSuperBox:
The addition of application menus made it more difficult to hit the target area for the messages indicator. Prior to the application menu addition, the entire blank area of the panel could be used. After, only the 2 gu space taken up by the icon could be used.

This was due to an implementation detail of the application menus... when the Shell has 60 gu or less width, both the application menus and the indicators are set to the full width of the panel. This meant that their drag handles, the area where they could be "pulled down" from, took the entire panel. Since the indicators were defined after the application menus in the file, the indicators' handle went over the application menus' and the menus were unusable.

To solve this, the original author forced the indicators' handle to only use the space taken by the indicator items. This approach had two big problems. First, if the user had enough indicators to reach all the way across the screen, the application menus would be unusable. Second, it shrank the usable handle area for the leftmost indicator.

This change adjusts the indicator bar to provide a 7 grid unit space to its left whenever the application menus are available. In case of space contention between the two (read: too many indicators are showing), the leftmost indicators will hide. Whenever the application menus are not available, the indicators (and, more importantly, their handle) take up the full width of the screen again.

Additionally, I added a "down" icon to the panel. It shows up at any screen size, but only when application menus are available. This should help improve discoverability of the application menus, which were visually "silent" in Staged mode prior to this change.

Indicators and panel sharing phone-sized area

Indicators and panel sharing very small area, showing indicators overflowing during space contention

Indicators and panel sharing a large area, showing window title appearing with "down" icon

Thanks @peat-psuwit for figuring out my mistakes and repainting over them to actually finish this PR.

Please rebase and merge or use a merge commit, the commit history is useful.

Fixes ubports/ubuntu-touch#1233
Fixes ubports/ubuntu-touch#1364


qml/Panel/Panel.qml
```diff
@@ -325,26 +328,53 @@ Item {
...
+            Label {
+                id: rowLabel
+                anchors {
+                    left: parent.left
+                }
+                objectName: "panelTitle"
+                width: root.width - __indicators.maxDragHandleSize - touchMenuIcon.width
+                height: root.minimizedPanelHeight
+                verticalAlignment: Text.AlignVCenter
+                elide: Text.ElideRight
```
mardy, line 351: Can the text be truncated? This property seems to imply that it can, but some other parts of the code make me think that the item is always being allocated the `implicitWidth` it requested.

qml/Panel/Panel.qml
```diff
@@ -325,26 +328,53 @@ Item {
...
+                visible: opacity !== 0
+                readonly property var implicitWidthWithMargin: implicitWidth === 0 ? 0 : implicitWidth + units.gu(1)
+            }
+
+            Icon {
+                id: touchMenuIcon
+                objectName: "touchMenuIcon"
+                anchors {
+                    right: parent.right
+                    rightMargin: parent.width - rowLabel.implicitWidthWithMargin - width
```
mardy: This code assumes that the `rowLabel` is always allocated the full `implicitWidthWithMargin`, but the `rowLabel`'s `width` property seems to be computed in some other way.

qml/Panel/Panel.qml
```diff
@@ -325,26 +328,53 @@ Item {
...
+                Behavior on opacity { NumberAnimation { duration: UbuntuAnimation.SnapDuration } }
+                visible: opacity !== 0
+                readonly property var implicitWidthWithMargin: implicitWidth === 0 ? 0 : implicitWidth + units.gu(1)
+            }
+
+            Icon {
+                id: touchMenuIcon
+                objectName: "touchMenuIcon"
+                anchors {
+                    right: parent.right
```
mardy: why isn't this anchored to the `rowLabel`'s right? Then the `rightMargin` property could be left unspecified.

qml/Panel/Panel.qml
```diff
@@ -37,8 +38,7 @@ Item {
 
     property real minimizedPanelHeight: units.gu(3)
     property real expandedPanelHeight: units.gu(7)
-    property real indicatorMenuWidth: width
-    property real applicationMenuWidth: width
+    property real menuWidth: partialWidth ? units.gu(40) : width
```
mardy, line 41: Am I understanding correctly, that with the old code, each of the menus could fill up the whole panel, whereas now, if the screen is large enough, each of them would be limited to 40 gu? So, if my screen would be 120 gus, only 40 gus on each side would be used, whereas the central 40 will always be empty?

mardy (changes requested):
I left a few comments, but mostly to understand how things work. Honestly, to me this code is mostly unreadable, and it looks terribly hard to maintain.

It may be that some of the things I'm suggesting are already implemented and I just failed to recognize them, but the way I'd proceed would be this:

1. Create a new item which has two children: the `__applicationMenus` panel and the `panelTitleHolder`. It will have an `implicitWidth` property set to the max of the desired widths of these two children, and when it gets its width set, it will enforce it on the children to make sure they won't escape it.

2. Add an `implicitWidth` property to the `__indicators` panel.

3. Make sure that the two items above (the new item and the `__indicators` one) expose all the properties that are needed to make a decision about their size and visibility.

4. Create a new `QtObject` that will take the two items above as properties and has the logic to compute their final width, and sets it on the items. Ideally (especially if this code is non trivial), move this QtObject to its own QML file, so that it will be much simpler to write unit tests for it.

5. In general, don't be shy in moving complex items into their own QML file.

As for the user experience itself, when there is not enough width to show the menus, I'd also shrink the indicator, but add a "..." symbol to make it clear that we are hiding some of them. And when the user starts a drag operation over the indicators area, they should get expanded and become all visible (maybe this is already the case).

qml/Panel/Panel.qml
UniversalSuperBox: image

oh.

UniversalSuperBox: image

This is fixed with the latest commit, as seen here. The gray rectangles are showing the handles which can be used to drag the menus down.

qml/Panel/Panel.qml
UniversalSuperBox: I believe this is fixed with the newest commit

qml/Panel/Panel.qml
UniversalSuperBox: Did exactly that

qml/Panel/Panel.qml
UniversalSuperBox: In the old code, there were no application menus. Therefore there was no need to make the indicators get out of the way of the application menus. Your assumption is correct, at least for the touch menus. No matter how wide your screen is, only 40gu is allocated for each menu.

image

(Shown with rectangles over the touch areas for clarity)

It's done this way because someone decided that the menu width should be the same as the handle width.

image

Global menus for mouse can take up more of the panel, and fade in when the mouse is over the panel. Similar to Unity7's approach.

I agree this is an issue, but since no known Ubuntu Touch apps use the application menus it should not be a priority right now.

qml/Panel/Panel.qml
```diff
@@ -48,6 +48,9 @@ Item {
     property bool greeterShown: false
     property bool hasKeyboard: false
 
+    // Whether our expanded menus should take up the full width of the panel
+    property bool partialWidth: width > units.gu(60)
```
peat-psuwit: This will become false when `width` is 1 pixel above 60 GUs. Is this intentional? Or, what is the correct approach here? Does it matter?

peat-psuwit (commented):
This is pretty large. So, here's my understanding:
- `PanelMenu`'s `overflowWidth` dictates how wide is the handle. However, this value can't be wider than the `width` itself.
- If handles overlap, whatever is on the top (declared later) get the touch.
- When the screen is small, both app menu and indicator bar take the whole screen width. However, indicator bar will retract `overflowWidth` by 7 GUs when app menu is enabled, enabling app menu to be interacted.
- However, when the screen is large (>60 GUs), each of them will take only 40 GUs. In this case, it seems like the 7 GU clearance is not given as it's not needed. The handles might still overlap, but the app menu is guaranteed to have at least 20 GUs to interact with.
- The window title is a separated entity, visible when the screen is large, and expand until the starts of indicator panel.

I have a pedantic comment below. But otherwise, LGTM but someone else must approve.

qml/Panel/Panel.qml
UniversalSuperBox: There's really no other way to make breakpoints. There will always be a one-pixel spot where you can cause the layout to change.

peat-psuwit: I mean, is this supposed to be `>=` instead of `>`? In other words, which one-pixel spot should we consider?

UniversalSuperBox: I'll use `>=`.

Flohack74:
@mardy can you do a quick summary review here, or should someone else take this over?

mardy:
> @mardy can you do a quick summary review here, or should someone else take this over?

Well, as I wrote [above](https://github.com/ubports/unity8/pull/289#pullrequestreview-372226669), there are several changes I'd suggest to make, but I don't think it should prevent this MR from being merged. Code-wise, it's not easy to understand, but on the other hand this is mostly due to the existing code -- this branch does only add a little drop to the ocean. :-)

I'd invite people from the community to test the PPA where this branch is built to and, if no regressions are found, just merge it.

UniversalSuperBox:
The changes to tests are much larger than the changes to production code, so this isn't quite as scary as " +522 −262" would have you believe.

Balancing mobile and desktop usability with the current reality of releasing OTA-12 and the requested refactoring's difficulty, I believe this should be merged.

To be more detailed about a refactoring, Mardy's requested changes aren't all that have been requested. @peat-psuwit also commented that a ListView inside a Flickable is a poor choice here, and leads to a lot of problems.

peat-psuwit (approved):
The new commit invalidates my review. So, again, LGTM, but someone else must approve. If @mardy agrees with me, please add another approval.

PR #291, opened 2020-03-10
Fix utils tests

UniversalSuperBox:
The old usage of the QCOMPARE() macro appears to have been removed between Qt 5.4 and 5.9. By creating and using a new variable we avoid issues.


NeoTheThird (approved):
Should be okay

PR #293, opened 2020-03-30
[DBusGreeterList] fix EntryIsLocked property

### LINKED ISSUE #1406: File Manager Authentication
<!File Manager authentication dialogue is now missing (Nexus 4 and BQ M10, both running devel 469). Same for Terminal. What is intention for the future?
(See recent discussion in Support channel.)--------------------------------------------------------------------------
--------------------------------------------------------------------------------

Thank you for wanting to file an issue! If you have a minute, please take the time to read our guide on writing a good bug-report:

    https://docs.ubports.com/en/latest/contribute/bugreporting.html

If not and you just want to get this over with, that's also quite alright. But please try to format your issue correctly, to make it easier for us to read. Below, we provide a template for a feature-request and a bug-report. Use common sense to determine what category your issue falls under and delete the template you don't need. Also, you don't need to stick to the templates under all circumstances. If you don't need a heading, remove it. If you think a new section will provide valuable information, add a new heading. We're not dogmatists, we're pragmatists who just want to be able to efficiently work with new issues.

--------------------------------------------------------------------------------
------------------------------------------------------------------------------->

- Device: Nexus 4 and BQ M10
- Channel: devel
- Build: 469 onwards

- Steps to reproduce
Launch File Manager

- Expected behavior
Authentication dialogue should be presented

- Actual behavior
No Authentication Dialogue appears and File Manager is already unlocked

peat-psuwit:
The property has been wired into Greeter's isAuthenticated(), in (merge)
commit 6d9907ef7fdb ("Use a model for PAM prompts, supporting more
possible interactions."). This changes its meaning; it's supposed to
mean whether the account has a lock (i.e. password), not whether the
screen is locked at the moment.

This commit brings back Greeter's promptless() property and using it for
EntryIsLocked, thus fixing the property's meaning.

A regression test has been added, and 2 tests are updated for the change
in change signaling order.

Fixes https://github.com/ubports/ubuntu-touch/issues/1406

### DIFF
--- a/plugins/LightDM/DBusGreeterList.cpp
+++ b/plugins/LightDM/DBusGreeterList.cpp
@@ -25,7 +25,7 @@ DBusGreeterList::DBusGreeterList(Greeter *greeter, const QString &path)
    m_greeter(greeter)
 {
     connect(m_greeter, &Greeter::authenticationUserChanged, this, &DBusGreeterList::authenticationUserChangedHandler);
-    connect(m_greeter, &Greeter::isAuthenticatedChanged, this, &DBusGreeterList::authenticatedChangedHandler);
+    connect(m_greeter, &Greeter::promptlessChanged, this, &DBusGreeterList::promptlessChangedHandler);
 }
 
 QString DBusGreeterList::GetActiveEntry() const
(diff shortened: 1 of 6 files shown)


Flohack74 (approved):
LGTM

UniversalSuperBox (approved):
Interestingly, it appears this property was not used outside of this very limited area. It may have been changed to make the language more clear, but apps were not changed with it. Until that can happen, this seems correct.

PR #294, opened 2020-03-30
Add pull-to-refresh to app drawer

peat-psuwit:
We have XdgWatcher that watches the change for Click apps. However,
there isn't such a thing for Libertine apps. This commit implement
pull-to-refresh in the app drawer as a stop-gap solution.

The refreshing action is synchronous and will block main thread.
Changing it to be asynchronous would involve changing AppDrawerModel's
interface more significantly and updating the test, something I'm
uncomfortable doing near OTA-12 release.

In the long term, when we have such mechanism, we can remove it. Or, we
can just leave it here as a failback when watcher things work
incorrectly.

|Pull to refresh|Release to refresh|
|----------------|-------------------|
|Pull to refresh|image|

### DIFF
--- a/plugins/Unity/Launcher/CMakeLists.txt
+++ b/plugins/Unity/Launcher/CMakeLists.txt
@@ -50,6 +50,6 @@ target_link_libraries(UnityLauncher-qml
     ${UAL_LIBRARIES}
     )
 
-qt5_use_modules(UnityLauncher-qml DBus Qml Gui)
+qt5_use_modules(UnityLauncher-qml DBus Qml Gui Concurrent)
 
 add_unity8_plugin(Unity.Launcher 0.1 Unity/Launcher TARGETS UnityLauncher-qml)
--- a/plugins/Unity/Launcher/appdrawermodel.cpp
+++ b/plugins/Unity/Launcher/appdrawermodel.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2016 Canonical, Ltd.
+ * Copyright (C) 2020 UBports Foundation.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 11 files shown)


UniversalSuperBox:
I believe this is an acceptable workaround for Libertine apps not refreshing properly on OTA-12.

However, do you know any way that it could be moved to a background thread? It tends to freeze the interface for a few seconds.

peat-psuwit:
Well, given the size of some of my other PRs, I might as well implement it asynchronously. So, yeah, I'll implement it.

peat-psuwit:
Turns out implementing asynchronous refresh is harder than I thought.

`AppDrawerModel` keeps its data in a `QList<LauncherItem *>`. So, the first idea is to re-generate this list in a worker thread. However, if the model is destructed before the worker finished refreshing, those `LauncherItem`s will become leaked. And, we can't pass the model's pointer in as `LauncherItem`'s parent because

1. It can become use-after-free. And
2. QObject prevents setting parent to an object in a different thread.

Thus, it's unavoidable to change the list to `QList<std::shared_ptr<...>>` instead.

UniversalSuperBox (approved):
No worries. Even if it's not perfect, the only time it needs to be used is when installing a Libertine app. In the future we'll need to fix either ubuntu-app-launch or the xdg-watcher to remedy the problem.

Bolly:
Checked this on a BQM10FHD running RC 2020 W-16.
It works.

ghost:
Works on my OnePlus 3t, too.

abkro:
Tested on N5 running RC 2020-W16 as follows:

* removed Geany, opened app drawer, pulled to refresh
* reinstalled Geany, repeated steps

It worked as expected in both cases.

j2g2rp:
works fine here on Nexus 5 and M10FHD:
Installed firefox, qalculate and geany. after refresh the apps are shown correctly in drawer.
Removed geany. After refresh the drawer geany launcher is not shown.

megahirtz:
Tested on a OnePlus One. Functions as expected adding and removing apps in Libertine containers.

PR #295, opened 2020-03-31
Show the Drawer tutorial for upgrading users

### LINKED ISSUE #280: Tutorial's "Swipe from left to see Drawer" step should run on upgrade from pre-Drawer Unity8
Some people have expressed confusion when they upgraded from pre-Drawer Unity8 to post-Drawer Unity8. The Drawer, while it shares the same gesture as the Dash, is not terribly discoverable.

We can help this situation by showing the Tutorial's "Long swipe from the left edge to open the Application Drawer", but not the rest of the Tutorial.

image

UniversalSuperBox:
With this change, users upgrading from Unity8 8.20 and below will receive the "Long swipe from the left to open the Drawer" tutorial. This should be the most important change that people used to Unity8 with the Dash will need to know, so we can show it to them right away.

Fixes https://github.com/ubports/unity8/issues/280


mariogrip (approved):
LGTM :+1:

PR #300, opened 2020-04-15
Repair localization: Bring in missing po updates from since ever

Flohack74:


UniversalSuperBox (approved):
CI Failure occurred only on arm64, so it's our usual randomly failing tests. Otherwise, despite the HUGE changeset, I guess we have to merge it to get translations back on track.

PR #314, opened 2020-05-26
WIP: greeter Theme changes

mateosalta:
try some theme changes

### DIFF
--- a/qml/Greeter/LoginAreaContainer.qml
+++ b/qml/Greeter/LoginAreaContainer.qml
@@ -33,6 +33,6 @@ Item {
     UbuntuShape {
         anchors.fill: parent
         aspect: UbuntuShape.Flat
-        backgroundColor: theme.palette.normal.raised
+        backgroundColor: theme.palette.normal.background
     }
 }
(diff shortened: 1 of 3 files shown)


mateosalta:
this removes the keyboard background, and changes hardcoded colors into theme colors, unfortunatly this part of lomiri is hardcoded to the dark theme. 

Will have to untangle the mess of specifying different themes in shell elements, then recolor items that we wish to stay a certain color.

UniversalSuperBox:
Are you still working on this @mateosalta?

mateosalta:
yeah, it's an ongoing effort to try and theme the lock-screen: https://forums.ubports.com/topic/6022/login-screen-code-input-design/105

It's getting there, but we went in a new direction, instead of opaque boxes we are going to try and use white elements like the initial lock-screen page with the time and infographics.

mateosalta:
unless I could get it to a intermediate stage, similar to this state: https://forums.ubports.com/assets/uploads/files/1619777417175-screenshot20210430_050857603.png but perhaps a shade on the emergency call as well, and then perform further iterations

- keyboard transparency works
- keep slight transparency on login box for now because of the error text (blocker for having no background box)

PR #324, opened 2020-08-09
Top panel from positive to activity

cibersheep:
Second try :)
- If we are in a call and Dialer is not visible, indicators bar shoud be activity and not positive.
- Let's check if we need to change Label color ~~https://github.com/ubports/unity8/blob/xenial/qml/Panel/Panel.qml#L356~~ from https://github.com/ubports/unity8/blob/xenial/qml/Panel/ActiveCallHint.qml#L76

### DIFF
--- a/qml/Panel/ActiveCallHint.qml
+++ b/qml/Panel/ActiveCallHint.qml
@@ -77,6 +77,7 @@ Item {
                 height: callHint.height
                 verticalAlignment: Text.AlignVCenter
                 text: i18n.tr("Tap to return to call...");
+                color: theme.palette.normal.activityText
             }
 
             Label {
@@ -94,6 +95,7 @@ Item {
                         return contactWatcher.alias !== "" ? contactWatcher.alias : contactWatcher.phoneNumber;
                     }
                 }
+                color: theme.palette.normal.activityText
             }
         }
     }
@@ -127,6 +129,7 @@ Item {
                     return m + ":0" + ss;
                 }
             }
+            color: theme.palette.normal.activityText
         }
 
         PathView {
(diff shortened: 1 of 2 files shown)


cibersheep:
Before:
imatge

After:
imatge

mariogrip:
I'm not sure if this make sense? since most people associate green (like the dialer button) with ongoing call. for me blue looks a bit off?

cibersheep:
I know what you mean but
- When we show a button to send an SMS we don't color it orange ;)
- The green is for positive actions. 
- The blue is to draw attention on something and for neutral actions. In my opinion, going back to Dialer it's not a positive nor destructive action

http://docs.ubports.com/humanguide/design-concepts/system-palette.html

giiba:
For what it is worth I think the blue looks really good.

kugiigi:
I think this is indeed debatable.
If I am not mistaken, green is also the usual color for calls in iOS and Android so it's good for familiarity. On the other hand, I think it does makes sense that the activity color (blue) is used since an on going call is an activity.

Personally, I'm leaning more towarss blue, I guess because it's my favorite color 😄

dobey:
From the conversation, and from my design experience, I'm not sure what we should do with this. But I would like to get it out of the PR queue either way.

So @cibersheep @UniversalSuperBox what should we do with this one?

cibersheep:
Green was the old color used for activity but moved to only positive actions. Blue for activity from 1.3 on.
So, in my opinion we should go ahead with this

UniversalSuperBox (approved):
I am willing to believe that this was green to match the old activity color, it changed without lomiri being updated. Let's go ahead.

PR #326, opened 2020-08-11
[Indicators] Add support for ayatana indicators

mariogrip:
This adds support for ayatana indicators but keeps the existing unity indicators to support legacy indicators, this also cleans up the indicator load logic.

This opts to add each indicator menu types to the qml object instead of doing a hacky replace string logic, this way we save cpu cycles but at a ultra minor memory bump.

### DIFF
--- a/plugins/Unity/Indicators/indicatorsmanager.cpp
+++ b/plugins/Unity/Indicators/indicatorsmanager.cpp
@@ -55,20 +55,22 @@ void IndicatorsManager::load()
 {
     unload();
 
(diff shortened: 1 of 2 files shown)


peat-psuwit (commented):
LGTM, but someone else must approve. Didn't test or anything.

As this merges to `xenial` branch, maybe wait for OTA-13 to be released first?

mariogrip:
@UniversalSuperBox what do you think?

dobey (changes requested):
Needs to update/add tests and data for these changes.

PR #328, opened 2020-08-24
 Add multi monitor and workspace support (for edge)

mariogrip:
DON'T merge this, this is an *proposal* to add into edge, but since we don't have an edge branch, i'll create it from this once this gets approved. IMO we should move this to edge to get some speed into the development of this, now that we have added devel rc and stable channels for arm64 i we can start to use edge to test this.     

This adds multi monitor and workspace support. Workspace is disabled by default, but can be forced on.

This is massive!

This also fixes many (if not all) of our mouse issues! YAY

Needs:
ubports/qtmir#25
ubports/unity-api#15


plugins/GlobalShortcut/globalshortcutregistry.cpp
```diff
@@ -102,24 +119,15 @@ bool GlobalShortcutRegistry::eventFilter(QObject *obj, QEvent *event)
     return QObject::eventFilter(obj, event);
 }
 
-void GlobalShortcutRegistry::setupFilterOnWindow(qulonglong wid)
+void GlobalShortcutRegistry::setupFilterOnWindow(QWindow* window)
 {
-    if (wid == s_windowId) {
```
mardy, line 107: Maybe replace this with
```
if (window == m_filteredWindow) {
    return;
}
```
or add an assertion, if this is impossible.

PR #339, opened 2020-10-20
tst_NarrowView: Use Qt's DoubleClick, not tapping

UniversalSuperBox:
Manually tapping twice in JavaScript was subject to some limitations in QML. In particular, only one tap event could happen per frame. It is possible for frames in the NarrowView test to take longer than the double-click interval, causing this sequence of taps to be misrecognized.

By using Qt's own mouseDoubleClickSequence, we avoid this limitation and the test succeeds no matter how slow the host is.


dobey (approved):
Looks good!

PR #349, opened 2020-12-22
Avoid jerky Drawer app list display while scrolling

### LINKED ISSUE #364: App drawer scroll is very slow
on BQ4.5 devel, the drawer app list scroll is jerky, can be very slow.

Please consider that work around : #349

lduboeuf:
The Drawer app list scroll can be jerky, especially on low-end devices. 
Found a work around to provide a faster scroll experience.

fixes #364

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -335,19 +335,23 @@ FocusScope {
                             fillMode: Image.PreserveAspectCrop
                         }
 
-                        UbuntuShape {
+                        OpacityMask {
+                            anchors.fill: sourceImage
                             source: sourceImage
-                            aspect: UbuntuShape.Flat
-                            width: parent.width
-                            height:parent.height
-                            radius : "medium"
-                            StyledItem {
-                                styleName: "FocusShape"
-                                anchors.fill: parent
-                                StyleHints {
-                                    visible: drawerDelegate.focused
-                                    radius: units.gu(2.55)
-                                }
+                            maskSource: UbuntuShape {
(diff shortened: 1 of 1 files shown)


lduboeuf:
humm, well in some cases it is smoother, in some other cases it can also be jerky. The issue here is the UbuntuShape which seems not very efficient.

cibersheep:
Could you add a screenshot? ^_^

lduboeuf:
Well a screenshot to show that the scroll is smoother is hard ;-)

cibersheep:
So, that means it still looks like it is now?
^___^

lduboeuf:
> So, that means it still looks like it is now?
> ^___^

yes no changes, just that the scroll is less jerky

lduboeuf:
for the rendering: 
screenshot20210128_232833558

UniversalSuperBox:
I find the use of an OpacityMask here very interesting -- what is it for? Does it change the rendering in any case?

lduboeuf:
> I find the use of an OpacityMask here very interesting -- what is it for? Does it change the rendering in any case?

It is a way to make images having a rounded corner by applying a mask from UbuntuShape. I don't see any difference in rendering but it get faster. Don't ask me why ;-)

UniversalSuperBox:
Interesting. There is no noticeable difference on any of my devices. On the OnePlus One, it appears even choppier with a long app list.

lduboeuf:
since 2 weeks, probably Mir changed ?, I no longer  have an issue with the Drawer

PR #355, opened 2021-01-12
Fixes for Qt 5.12 migration

dobey:
Would like to have cleaner commit history with some stuff squashed together more, but alas.


debian/control
```diff
@@ -2,214 +2,224 @@ Source: unity8
 Section: x11
 Priority: optional
 Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
-Build-Depends: cmake,
```
peat-psuwit, line 5: You created the indentation change in the same commit as another change. This makes it hard to find out what's been changed here.

plugins/Utils/windowstatestorage.cpp
```diff
@@ -171,16 +156,13 @@ void WindowStateStorage::initdb()
 void WindowStateStorage::saveValue(const QString &queryString)
 {
     QMutexLocker mutexLocker(&s_mutex);
-
-    QFuture<void> future = QtConcurrent::run(&m_threadPool, executeAsyncQuery, queryString);
```
peat-psuwit, line 175: Hmm, I'm not sure it's a good idea to do synchronous execution. Does the test just save the state and then read it back immediately? We may let the test retry, or we may make the get function waits for remaining queries before getting the value. The get function is synchronous though...

tests/qmltests/Launcher/tst_Launcher.qml
```diff
@@ -941,7 +942,7 @@ Rectangle {
             tryCompare(quickListShape, "opacity", 0.95);
             mouseRelease(draggedItem);
 
-            tryVerify(function() {return (quickList.y >= units.gu(1))});
+            tryVerify(function() {return (quickList.y >= units.gu(0.5))});
```
UniversalSuperBox, line 959: Why has the quickList's y value changed?

tests/qmltests/tst_OrientedShell.qml
```diff
@@ -914,7 +916,7 @@ Rectangle {
             compare(twitterApp.supportedOrientations, Qt.PortraitOrientation | Qt.LandscapeOrientation
                     | Qt.InvertedPortraitOrientation | Qt.InvertedLandscapeOrientation);
 
-            tryCompare(twitterDelegate, "orientationChangesEnabled", true);
+            tryCompare(twitterDelegate, "orientationChangesEnabled", false);
```
UniversalSuperBox: Why has this value changed? I mean, it seems more correct the new way, but what's different?

tests/qmltests/tst_ShellWithPin.qml
```diff
@@ -542,9 +550,6 @@ Item {
             tap(shell);
             tryCompare(launcher, "state", "");
 
-            console.log("I AM LOGGING NOW");
-            console.log(ApplicationManager.count);
```
UniversalSuperBox, line 546: uh... I can explain

tests/mocks/Unity/Application/SurfaceManager.cpp
```diff
@@ -35,6 +35,9 @@ SurfaceManager *SurfaceManager::m_instance = nullptr;
 
 SurfaceManager *SurfaceManager::instance()
 {
+    if (m_instance == nullptr) {
+        auto inst = new SurfaceManager();
+    }
```
UniversalSuperBox, line 40: This causes a warning as the compiler thinks it is a useless statement. Obviously it isn't, as it ends up setting m_instance as a side effect.

In more interesting news, somewhere around here is the cause of the failures in the tst_ShellWithPin tests. When built with debugging enabled, the Q_ASSERT below (`Q_ASSERT(m_instance == nullptr)`) fails:

```
QFATAL : qmltestrunner::ShellWithPin::test_bfbOnLockedDevice() ASSERT: "m_instance == nullptr" in file /home/ubuntu/unity8/tests/mocks/Unity/Application/SurfaceManager.cpp, line 45
FAIL!  : qmltestrunner::ShellWithPin::test_bfbOnLockedDevice() Received a fatal error.
```

It appears that we may have two instances of a SurfaceManager in flight during the tests... I'm not sure why that would be, though. Maybe it's okay to have more than one instance of a SurfaceManager and the instance shouldn't be static. Maybe we need to do something to ensure that the old one is destroyed.

UniversalSuperBox (changes requested):

tests/mocks/Unity/Application/SurfaceManager.cpp
UniversalSuperBox: Wait, resync. The test failure when *not* built with debugging is the new application never gets a Surface, leading to the wrong thing being focused. So having two SurfaceManagers definitely wreaks havoc.

qml/Launcher/LauncherDelegate.qml
```diff
@@ -37,7 +37,7 @@ Item {
 
     readonly property int effectiveHeight: Math.cos(angle * Math.PI / 180) * itemHeight
     readonly property real foldedHeight: Math.cos(maxAngle * Math.PI / 180) * itemHeight
-    readonly property alias wiggling: wiggleAnim.running
+    readonly property alias wiggling: root.alerting
```
peat-psuwit: Does the old code cause any problems? It seems to be written this way to make sure this property is changed after the animation is actually running. See https://github.com/ubports/unity8/commit/e46dc1dc552f1ddbc27de581b49ba5dfda877db4.

tests/qmltests/Stage/tst_PhoneStage.qml
```diff
@@ -251,10 +251,11 @@ Item {
...
             for (var i = 0; i < appsToTest; i++) {
-                var item = new Object();
-                item.tag = "App " + i;
-                item.index = i;
-                item.total = appsToTest;
+                var item = {
+                    tag: "App " + i,
+                    index: i,
+                    total: appsToTest
+                };
```
peat-psuwit, line 258: Although I actually prefer this, why does the old code cause an error?

tests/mocks/Unity/Application/SurfaceManager.cpp
peat-psuwit: There must be the code somewhere that's supposed to create an instance of this class, be it C++ or QML. In what test does this cause a failure?

qml/Tutorial/TutorialContent.qml
```diff
@@ -77,15 +77,15 @@ Item {
 
         readonly property bool skipped: !d.mobileScenario || d.haveShown("top")
         readonly property bool shown: item && item.shown
-        active: !skipped || (item && item.visible)
+        active: !skipped
```
peat-psuwit, line 80: What does this fix?

peat-psuwit (changes requested):

qml/Tutorial/TutorialContent.qml
dobey: Some `TypeErrors`

tests/qmltests/Stage/tst_PhoneStage.qml
dobey: I think it was related to the `Object` issues from uitk, as there was a `var item = new Object();` instead.

qml/Launcher/LauncherDelegate.qml
dobey: IIRC, it was a binding loop, since `wiggleAnim.running` just points at `alerting`.

plugins/Utils/windowstatestorage.cpp
dobey: IIRC, it was just crashing.

debian/control
dobey: It was hard to follow before and needed to adjust things while building, so yeah, it's an unfortunate side effect.

tests/qmltests/tst_OrientedShell.qml
dobey: It seems more correct given the other comparisons in the test, and I don't know why it wasn't failing before or why it started failing when building against Qt 5.12, exactly.

tests/qmltests/Launcher/tst_Launcher.qml
dobey: I'm not sure why exactly it is different in the tests, but when opening the quick list in a live system, it is positioned correctly.

plugins/Utils/windowstatestorage.cpp
mariogrip: Yeah this should def be kept as async. If you want I can look into this.

tests/mocks/Unity/Application/SurfaceManager.cpp
mariogrip: Something should be creating this in the first place, so where does it crash?

Other point: Well the compiler is correct, this creates a unused variable, `inst` is not used anywhere. No need to create a var, you can just do `new SurfaceManager();` without setting a var.

tests/qmltests/tst_ShellWithPin.qml
```diff
@@ -282,6 +282,7 @@ Item {
         }
 
         function test_emergencyCallCrash() {
+            skip("Failing on Qt 5.12 with other tests skipped");
```
mariogrip: Lots of skipping here, lets create a issue to fix these.

tests/qmltests/tst_OrientedShell.qml
mariogrip: This is correct? orientationChangesEnabled should still be true as we compere supportedOrientations above and the source code says if supportedOrientations = PortraitOrientation && LandscapeOrientation then supportedOrientations should be true

See:
https://github.com/ubports/unity8/blob/xenial/qml/Stage/ApplicationWindow.qml#L28
https://github.com/ubports/unity8/blob/xenial/qml/Stage/ApplicationWindow.qml#L105

qml/Tutorial/TutorialContent.qml
mariogrip: Would this not mean if Tutorial still ran even if its "skipped" it would make it inactive? I assume it is there to keep it active when its shown regardless if skipped is true.

qml/Components/KeyboardShortcutsOverlay.qml
```diff
@@ -198,30 +198,6 @@ Rectangle {
...
-                Label {
-                    Layout.columnSpan: 2
-                    text: i18n.tr("Scopes")
-                    font.weight: Font.Light
-                    color: theme.palette.normal.baseText
-                    lineHeight: 1.3
-                }
-
-                Label {
-                    text: i18n.tr("Super (Tap)")
```
mariogrip, line 214: Super (Tap) opens the side "applauncher" maybe we should keep this but change from scope to App Launcher?

mariogrip (changes requested):

tests/mocks/Unity/Application/SurfaceManager.cpp
UniversalSuperBox: No, there is only ever one SurfaceManager instance at a time in the tests. It's a static variable, which is why the assertion to check there is only one of them exists.

I've investigated what it would take to make SurfaceManager::instance() a non-static object, which would be best, but unfortunately a lot of other mocks depend on its stasis. We'd need to implement much more of the Mir stack the same way it's implemented in production to fix that... And once that's done, there are other mocks implemented in this static way.

So, we need to ensure that there is only ever one SurfaceManager in flight. Why isn't there one now?

The answer is... race conditions. QML object destruction is unordered and asynchronous, so in the tests (specifically in ShellWithPin), it's possible to destroy one Shell instance but create a new one before all of the old one's child objects are removed. If the SurfaceManager is the object that happens to remain in flight... boom.

Right now I'm trying to focus on a way to ensure the old object tree is deleted before the new one is instantiated. Turns out that's really difficult... If you call `.destroy()` on an object, the deletion is queued for whenever the event loop is ready to do it. Even running `gc()` manually doesn't help for some unholy reason.

So... now I'm working on using [TestCase.createTemporaryObject](https://doc.qt.io/qt-5/qml-qttest-testcase.html#createTemporaryQmlObject-method), which seems to be the intended way to solve this problem.

tests/qmltests/Stage/tst_PhoneStage.qml
peat-psuwit: Right.

qml/Tutorial/TutorialContent.qml
dobey: In what cased would it ever be skipped and still visible? AFAICT, it should only ever be visible if it's not skipped and wasn't already completed. I don't see why it would ever be skipped and still visible.

tests/mocks/Unity/Application/SurfaceManager.cpp
dobey: I've pulled in the changes from @UniversalSuperBox related to this, so I suppose this can be resolved now?

tests/qmltests/tst_ShellWithPin.qml
UniversalSuperBox: These have been fixed

qml/Launcher/LauncherDelegate.qml
UniversalSuperBox: No, but the test does fail if the old value is used. I think it should be using a SignalSpy anyway, so I'll get that set up...

UniversalSuperBox: Nope! There's actually a bug here. Wiggling only starts on the first time the count becomes visible. `alerting` is never cleared so the wiggle can start the second time.

UniversalSuperBox: It appears that the normal LauncherModel never clears the `alerting` property when the count becomes invisible, either. I think this is expected behavior and the test should be resetting the property between runs.

UniversalSuperBox: All of those hypotheses were wrong. The problem is actually that we're using [alwaysRunToEnd](https://doc.qt.io/qt-5/qml-qtquick-animation.html#alwaysRunToEnd-prop), meaning the animation can be stopped (which sets `running` to `false`) before it actually completes and visually stops.

UniversalSuperBox: Addressed in https://github.com/ubports/unity8/pull/355/commits/627bd76e70ff6e3d5ab427a7103d8e0039be8c2e

tests/qmltests/tst_OrientedShell.qml
UniversalSuperBox: I believe the new value is more correct than the old value.

dobey: After the other fixes for not using a Loader for the shell in the tests, this seems to work correctly again, so reverted.

tests/mocks/Unity/Application/SurfaceManager.cpp
mariogrip: Should this not be removed then?

qml/Components/KeyboardShortcutsOverlay.qml
dobey: I don't think we need to include it (and I'd prefer to not introduce new strings here). If we feel it needs to be included, we can add it in a later commit.

PR #356, opened 2021-01-18
Rewrite WindowStateStorage for improved async power

UniversalSuperBox:
6931e29 changed WindowStateStorage from asynchronous writes to synchronous due to changes in QSql that denied reusing connections in different threads. This commit splits out all QSql functionality from the WindowStateStorage object into a separate AsyncQuery object and moves the AsyncQuery object into another thread. This allows performing write operations in another thread, preventing main thread freezes from slow IO.

If database initialization fails, the object simply returns the default values provided and black-holes writes.

Also adds tests to ensure the database name is being used sanely and that SQL injection is guarded against.

Co-Authored-By: Rodney Dawes <dobey@ubports.com>

### DIFF
--- a/plugins/Utils/windowstatestorage.cpp
+++ b/plugins/Utils/windowstatestorage.cpp
@@ -275,7 +275,7 @@ WindowStateStorage::WindowStateStorage(const QString &dbName, QObject *parent):
 {
     qRegisterMetaType<WindowStateStorage::WindowState>("WindowStateStorage::WindowState");
     QString dbFile;
-    if (dbName == nullptr) {
+    if (dbName.isEmpty()) {
         const QString dbPath = QStandardPaths::writableLocation(QStandardPaths::GenericCacheLocation) + QStringLiteral("/unity8/");
         QDir dir;
         dir.mkpath(dbPath);
--- a/plugins/Utils/windowstatestorage.cpp
+++ b/plugins/Utils/windowstatestorage.cpp
@@ -48,17 +48,23 @@ public:
 
     Q_INVOKABLE const QString getDbName()
     {
+        if (!m_ok) {
+            return "ERROR";
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         return connection.databaseName();
     }
 
     Q_INVOKABLE bool initdb()
     {
+        if (m_ok) {
+            return true;
+        }
         QSqlDatabase connection = QSqlDatabase::addDatabase(QStringLiteral("QSQLITE"), m_connectionName);
         connection.setDatabaseName(m_dbName);
         connection.setConnectOptions(QStringLiteral("QSQLITE_BUSY_TIMEOUT=1000"));
         if (!connection.open()) {
-            qWarning() << "AsyncQuery::initdb: Error opening state database " << m_dbName << connection.lastError().driverText() << connection.lastError().databaseText();
+            qWarning() << "AsyncQuery::initdb: Error opening state database. Window positions will not be saved or restored." << m_dbName << connection.lastError().driverText() << connection.lastError().databaseText();
             return false;
         }
         QSqlQuery query(connection);
@@ -86,11 +92,15 @@ public:
                 return false;
             }
         }
+        m_ok = true;
         return true;
     }
 
     Q_INVOKABLE int getState(const QString &windowId) const
     {
+        if (!m_ok) {
+            return -1;
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_getStateQuery);
@@ -116,6 +126,9 @@ public:
 
     Q_INVOKABLE QRect getGeometry(const QString &windowId) const
     {
+        if (!m_ok) {
+            return QRect();
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_getGeometryQuery);
@@ -154,6 +167,9 @@ public:
 
     Q_INVOKABLE int getStage(const QString &appId) const
     {
+        if (!m_ok) {
+            return -1;
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_getStageQuery);
@@ -181,6 +197,9 @@ public Q_SLOTS:
 
     void saveState(const QString &windowId, WindowStateStorage::WindowState state)
     {
+        if (!m_ok) {
+            return;
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_saveStateQuery);
@@ -193,6 +212,9 @@ public Q_SLOTS:
 
     void saveGeometry(const QString &windowId, const QRect &rect)
     {
+        if (!m_ok) {
+            return;
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_saveGeometryQuery);
@@ -208,6 +230,9 @@ public Q_SLOTS:
 
     void saveStage(const QString &appId, int stage)
     {
+        if (!m_ok) {
+            return;
+        }
         QSqlDatabase connection = QSqlDatabase::database(m_connectionName);
         QSqlQuery query(connection);
         query.prepare(m_saveStageQuery);
@@ -227,6 +252,7 @@ private:
     const QString m_getStageQuery = QStringLiteral("SELECT stage FROM stage WHERE appId = :appId");
     const QString m_saveStageQuery = QStringLiteral("INSERT OR REPLACE INTO stage (appId, stage) values (:appId, :stage)");
     QString m_dbName;
+    bool m_ok = false;
 
     void logSqlError(const QSqlQuery query) const
     {
@@ -254,20 +280,14 @@ WindowStateStorage::WindowStateStorage(const QString& dbName, QObject *parent):
     m_asyncQuery->moveToThread(&m_thread);
     connect(&m_thread, &QThread::finished, m_asyncQuery, &QObject::deleteLater);
     m_thread.start();
-    bool queryInitSuccessful;
+    // Note that we're relying on initdb being called before any other methods
+    // on AsyncQuery. Given the current behavior of QueuedConnection (slots
+    // invoked in the order they are received), this is fine.
     QMetaObject::invokeMethod(m_asyncQuery, "initdb",
-                              Qt::BlockingQueuedConnection,
-                              Q_RETURN_ARG(bool, queryInitSuccessful));
-    if (!queryInitSuccessful) {
-        qWarning() << "WindowStateStorage Failed to initialize AsyncQuery! Windows will not be restored to their previous location.";
-        m_asyncOk = false;
-        m_thread.quit();
-    } else {
-        connect(this, &WindowStateStorage::saveState, m_asyncQuery, &AsyncQuery::saveState);
-        connect(this, &WindowStateStorage::saveGeometry, m_asyncQuery, &AsyncQuery::saveGeometry);
-        connect(this, &WindowStateStorage::saveStage, m_asyncQuery, &AsyncQuery::saveStage);
-        m_asyncOk = true;
-    }
+                              Qt::QueuedConnection);
+    connect(this, &WindowStateStorage::saveState, m_asyncQuery, &AsyncQuery::saveState);
+    connect(this, &WindowStateStorage::saveGeometry, m_asyncQuery, &AsyncQuery::saveGeometry);
+    connect(this, &WindowStateStorage::saveStage, m_asyncQuery, &AsyncQuery::saveStage);
 }
 
 WindowStateStorage::~WindowStateStorage()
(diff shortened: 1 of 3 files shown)


plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,319 @@
...
 
-    QSqlQuery query = getValue(queryString);
+void WindowStateStorage::saveState(const QString &windowId, WindowStateStorage::WindowState windowState)
+{
+    if (!m_asyncOk) {
+        return;
+    }
+    // This could be a simple connection like all the other save* methods
+    // on AsyncQuery, but WindowStateStorage::WindowState can't be used
+    // on it directly.
```
UniversalSuperBox: I think there should be a way to use WindowStateStorage::WindowState as an argument to AsyncQuery::saveState, but I haven't been able to find it yet. It compiles if I flip the order of AsyncQuery and WindowStateStorage in the header file and forward-declare AsyncQuery, but the connect call fails saying that WindowStateStorage::WindowState is not a registered metatype.

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,319 @@
...
+AsyncQuery::AsyncQuery(const QString& dbName):
+    m_dbName(dbName)
+{
+}
 
-inline QString sanitiseString(QString string) {
-    return string.remove(QLatin1Char('\"'))
-                 .remove(QLatin1Char('\''))
-                 .remove(QLatin1Char('\\'));
+AsyncQuery::~AsyncQuery()
```
UniversalSuperBox: Should AsyncQuery be in its own \_p header file? How do I do that correctly?

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,319 @@
...
-    auto ok = query.exec(queryString);
-    if (!ok) {
-        qWarning() << "Error executing query" << queryString
-                   << "Driver error:" << query.lastError().driverText()
-                   << "Database error:" << query.lastError().databaseText();
+    if (!m_asyncOk) {
+        return defaultValue;
     }
+    QRect geometry;
+    QMetaObject::invokeMethod(m_asyncQuery, "getGeometry", Qt::BlockingQueuedConnection,
```
UniversalSuperBox, line 341: I have a feeling that a read encountering a write in progress would freeze the UI due to the database being locked. Should we time out the queries to prevent that?

plugins/Utils/windowstatestorage.cpp
mariogrip: Imo we should just place it in on top of the cpp file, it is not shared anywhere/

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,319 @@
...
+{
+}
 
-inline QString sanitiseString(QString string) {
-    return string.remove(QLatin1Char('\"'))
-                 .remove(QLatin1Char('\''))
-                 .remove(QLatin1Char('\\'));
+AsyncQuery::~AsyncQuery()
+{
+    QSqlDatabase::removeDatabase(m_connectionName);
```
mariogrip: Is this always safe to run? even if there is no database to remove?

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,319 @@
...
+    } else {
+        dbFile = dbName;
+    }
+    m_asyncQuery = new AsyncQuery(dbFile);
+    m_asyncQuery->moveToThread(&m_thread);
+    connect(&m_thread, &QThread::finished, m_asyncQuery, &QObject::deleteLater);
+    m_thread.start();
+    bool queryInitSuccessful;
+    QMetaObject::invokeMethod(m_asyncQuery, "initdb",
+                              Qt::BlockingQueuedConnection,
```
mariogrip: Does this need to be blocking? And if it some of the funcs get called before its inited, maybe use QWaitCondition or something in the funcs to make sure its inited,

plugins/Utils/windowstatestorage.h
```diff
@@ -47,25 +84,25 @@ class WindowStateStorage: public QObject
...
-    static QMutex s_mutex;
+Q_SIGNALS:
+    void saveStage(const QString &appId, int stage);
+    void saveGeometry(const QString &windowId, const QRect &rect);
 
-    // NB: This is accessed from threads. Make sure to mutex it.
-    QSqlDatabase m_db;
+private:
+    QThread m_thread;
+    AsyncQuery *m_asyncQuery;
```
mariogrip, line 73: Would like to see a std::shared_ptr here, else you need to make sure it gets destroyed after the class is deconstructed. but we should avoid using raw ptrs. (even if we use .get())

mariogrip (dismissed):
Some small things

plugins/Utils/windowstatestorage.cpp
UniversalSuperBox: Yes.

plugins/Utils/windowstatestorage.h
UniversalSuperBox: See https://github.com/ubports/unity8/pull/356/files#diff-95c2b40a9d9f8f86e9324b7ded598588aebd8ef27d61c36021d3e538e492e0c9R232, which will ensure the object is deleted when the thread stops.

plugins/Utils/windowstatestorage.cpp
UniversalSuperBox: Well, I suppose it would make Lomiri's start a little faster under conditions of high IO load... but the first application that the user launched would still have to wait on the database to initialize. In other words, you're just moving the delay to first app launch instead of Lomiri launch. Is that worth it?

plugins/Utils/windowstatestorage.cpp
UniversalSuperBox: I have moved AsyncQuery out of the header file. It is now declared in the .cpp.

plugins/Utils/windowstatestorage.cpp
UniversalSuperBox: I got it, but I'm not sure if I did it the best way possible. Where should I put the qRegisterMetaType call? https://github.com/ubports/unity8/pull/356/commits/02ca0715eae67635a597bd1ced218107a5b21e86

plugins/Utils/windowstatestorage.cpp
UniversalSuperBox: I added a timeout in https://github.com/ubports/unity8/pull/356/commits/209b85132073915f87ba111155056e05d751ec89

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,329 @@
...
+                              Q_RETURN_ARG(bool, queryInitSuccessful));
+    if (!queryInitSuccessful) {
+        qWarning() << "WindowStateStorage Failed to initialize AsyncQuery! Windows will not be restored to their previous location.";
+        m_asyncOk = false;
+        m_thread.quit();
```
peat-psuwit: We may want to nullify `m_asyncQuery` here (and then `m_asyncOk` isn't really needed), unless you decided to implement @mariogrip's comment at line 259 (see my comment).

plugins/Utils/windowstatestorage.cpp
peat-psuwit: Hmm, it can be some time between this class is init and the user starts an app, so it could be an overall win. I'm not sure about this though, and the complexity factor might make this idea not worth it.

peat-psuwit (commented):
The thing is, by creating a new `QSqlDatabase` every time, we don't get the benefit of having a dedicated thread, because Qt closes the DB connection as soon as the last `QSqlDatabase` is gone. (Edit: this turns out to be incorrect, see the comment below)

You should be able to just keep `QSqlDatabase` as an `AsyncQuery`'s member, and then replace it with a default-constructed instance before calling `QSqlDatabase::removeDatabase()` in the destructor.

Also...

UniversalSuperBox:
From https://doc.qt.io/qt-5/qsqldatabase.html#details (emphasis mine):

> Warning: It is highly recommended that you **do not keep a copy of the QSqlDatabase around as a member of a class**, as this will prevent the instance from being correctly cleaned up on shutdown. **If you need to access an existing QSqlDatabase, it should be accessed with database()**. If you chose to have a QSqlDatabase member variable, this needs to be deleted before the QCoreApplication instance is deleted, otherwise it may lead to undefined behavior.

I don't see any evidence that the connection is closed every time the QSqlDatabase object goes out of scope. Rather, the connection, which is global to the application, is kept open. The `QSqlDatabase` class is only an accessor to QSql's application-global backend.

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,329 @@
...
+    {
+        QSqlDatabase::removeDatabase(m_connectionName);
+    }
 
-void WindowStateStorage::saveState(const QString &windowId, WindowStateStorage::WindowState state)
-{
-    const QString queryString = QStringLiteral("INSERT OR REPLACE INTO state (windowId, state) values ('%1', '%2');")
-            .arg(sanitiseString(windowId))
-            .arg((int)state);
+    Q_PROPERTY (const QString dbName READ getDbName)
```
mardy: minor: typically properties are declared right below the `Q_OBJECT` macro (whether it's private or public, it does not matter)

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,329 @@
...
 
-    QSqlQuery query = getValue(queryString);
+private:
+    const QString m_connectionName = QStringLiteral("WindowStateStorage");
+    const QString m_getStateQuery = QStringLiteral("SELECT state FROM state WHERE windowId = :windowId");
+    const QString m_saveStateQuery = QStringLiteral("INSERT OR REPLACE INTO state (windowId, state) values (:windowId, :state)");
+    const QString m_getGeometryQuery = QStringLiteral("SELECT * FROM geometry WHERE windowId = :windowId");
+    const QString m_saveGeometryQuery = QStringLiteral("INSERT OR REPLACE INTO geometry (windowId, x, y, width, height) values (:windowId, :x, :y, :width, :height)");
+    const QString m_getStageQuery = QStringLiteral("SELECT stage FROM stage WHERE appId = :appId");
+    const QString m_saveStageQuery = QStringLiteral("INSERT OR REPLACE INTO stage (appId, stage) values (:appId, :stage)");
```
mardy: all these can be made `static`

peat-psuwit:
I mean, I see this statement from [the destructor's document](https://doc.qt.io/qt-5/qsqldatabase.html#dtor.QSqlDatabase):
> Note: When the last connection is destroyed, the destructor implicitly calls close() to release the database connection.

However, reading the Qt code, apparently there's a `QSqlDatabase` instance in a global cache that will get destroyed when `removeDatabase()` is called. So, this statement, while technically true, is confusing (and misleading). I should file a Qt bug at some point...

plugins/Utils/windowstatestorage.cpp
mardy: A Queued connection is enough, really:

> It needs to be synchronous, else it is possible to save a value before the connection is opened. We don't really want that.

This cannot happen, because the thread is running an event loop, and will process all the queued requests in the same order they were sent: so it will always process the `initdb()` method first.

plugins/Utils/windowstatestorage.cpp
mardy: Wait... AFAIK, in SQLite a reader is never blocked; the only operation that can be blocked is the write one.

UniversalSuperBox: In section 5.0 of [File Locking and Concurrency in SQLite Version 3](https://sqlite.org/lockingv3.html), it is specified that there is a short period of time during a write that readers are blocked. So it's totally possible, and it was cheap enough to add the timeout.

plugins/Utils/windowstatestorage.h
mardy: Yes, here using deleteLater is safer, since the object is in another thread

mardy (approved):

plugins/Utils/windowstatestorage.h
```diff
@@ -44,28 +50,25 @@ class WindowStateStorage: public QObject
     Q_DECLARE_FLAGS(WindowStates, WindowState)
     Q_FLAG(WindowStates)
 
-    WindowStateStorage(const QString& dbName = nullptr, QObject *parent = nullptr);
+    WindowStateStorage(const QString &dbName = nullptr, QObject *parent = nullptr);
```
peat-psuwit: One is not supposed to assign or use `nullptr` with a reference. In this case, use an empty `QString()`. This is actually @dobey's fault, but let's fix it here altogether.

```suggestion
    WindowStateStorage(const QString &dbName = QString(), QObject *parent = nullptr);
```

plugins/Utils/windowstatestorage.cpp
```diff
@@ -16,167 +17,344 @@
...
+const QString AsyncQuery::m_getStageQuery = QStringLiteral("SELECT stage FROM stage WHERE appId = :appId");
+const QString AsyncQuery::m_saveStageQuery = QStringLiteral("INSERT OR REPLACE INTO stage (appId, stage) values (:appId, :stage)");
+
+WindowStateStorage::WindowStateStorage(const QString &dbName, QObject *parent):
+    QObject(parent),
+    m_thread()
+{
+    qRegisterMetaType<WindowStateStorage::WindowState>("WindowStateStorage::WindowState");
+    QString dbFile;
+    if (dbName == nullptr) {
```
peat-psuwit:
```suggestion
    if (dbName.isEmpty()) {
```

peat-psuwit (changes requested):

dobey (approved):
Looks OK.

peat-psuwit (approved):
Should be Ok now.

PR #358, opened 2021-01-22
build: migrate CoverageReport to cmake-extras & fix its usage

peat-psuwit:
Uses CoverageReport from cmake-extras and moves its usage to the correct place. This should allow the coverage-reporting build to be built. At least in theory; I tested only the configure phase and didn't actually compile or generate the actual report out of it.


dobey (changes requested):
Please also remove the coverage related files from the `cmake/modules/` directory.

peat-psuwit:
Oh, Unity8 includes its own copy of `EnableCoverageReport.cmake`? That's why I didn't see the deprecation message...

So, this turns out to be migrating from the vendored copy to the copy in cmake-extras. And look like the copy in this package has not been updated since the project was first committed, so I'm not sure if the usage is still compatible.

Given that the motivation of this PR is to make it possible to bring the newest cmake-extras into xenial so that mir-android-platform can use its GTest versioning, and my ultimate decision to not use that facility in the end, I'm not sure if it's worth it to check that.

dobey:
Yes, there's a local copy left over from when various projects simply included their own copies of things, and we didn't have `cmake-extras` package yet. Everything else still works the same, so it's safe to remove the vendored files for it, and simply depend on cmake-extras for this.

CMakeLists.txt
```diff
@@ -190,6 +175,22 @@ else()
...
+#  * Find html report in subdir coveragereport
+#  * Find xml report feasible for jenkins in coverage.xml
+#####################################################################
+if(cmake_build_type_lower MATCHES coverage)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --coverage" )
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} --coverage" )
+  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} --coverage" )
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} --coverage" )
+  ENABLE_COVERAGE_REPORT(TARGETS ${SHELL_APP} FILTER /usr/include ${CMAKE_SOURCE_DIR}/tests/* ${CMAKE_BINARY_DIR}/*)
+endif()
```
dobey, line 192: Actually, only the `ENABLE_COVERAGE_REPORT` call needs to remain. The `cmake-extras` module will automatically add the `--coverage` arguments. So we can get rid of the extra stuff.

dobey (changes requested):

CMakeLists.txt
peat-psuwit: Seems like `ENABLE_COVERAGE_REPORT` also has the `TESTS` parameter. Am I supposed to put anything?

dobey: Ah, yes. It should list the tests targets. I guess we should refactor it a lot more to bubble up all the targets in a couple variables, from all the subdirs.

PR #359, opened 2021-01-23
fix for notification led

wdehoog:
It seems due to the Qt 5.12 update the binding on hasMessages stopped working (why?).

The binding on icons seems to work fine so let's use that. See https://github.com/ubports/ubuntu-touch/issues/1651.

No idea how this could pass the tests but fail on the device.

Also no idea if these changes pass the tests. Lets see...

### DIFF
--- a/qml/Panel/Indicators/IndicatorsLight.qml
+++ b/qml/Panel/Indicators/IndicatorsLight.qml
@@ -187,9 +187,8 @@ console.log("no support for Multicolor LED. " + indicatorState)
         actionGroup: _actionGroup
         actionName: "messages"
         Component.onCompleted: actionGroup.start()
-        property bool hasMessages: valid && (String(icons).indexOf("indicator-messages-new") != -1)
-        onHasMessagesChanged: {
-            root.hasMessages = hasMessages
+        onIconsChanged: {
+            root.hasMessages = valid && String(icons).indexOf("indicator-messages-new") > -1
             updateLightState("onHasMessagesChanged")
         }
     }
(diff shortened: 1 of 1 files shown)


dobey (changes requested):
While this might work around the issue, an even simpler workaround is to simply remove `valid` from the `hasMessages` check above. It seems there is where the problem actually lies, and it is not being modified properly from the C++ for some reason.

wdehoog:
To me the ``valid`` property seems needed. If I am correct it is determined here: https://github.com/ubports/unity8/blob/ac69b162683ecd80d4bcfe877d4444797f1f6930/plugins/Unity/Indicators/actionrootstate.cpp#L77

Since the icon names are checked I think it is also better to use the onIconsChanged event.

dobey:
> To me the `valid` property seems needed.

Then I guess we should figure out why `validChanged()` is not being called, and fix that, no? This is the reason that `hasMessages` isn't changed either. For some reason `validChanged` is not being emitted, thus Qt doesn't reevaluate the `hasMessages` binding to it.

wdehoog:
Even if ``validChanged`` is not fired (which indeed isn't) it's value can still be checked.

Removing ``valid`` does indeed make it work. Putting ``valid`` at the end of the expression also makes it work.
How can this happen?

dobey:
> How can this happen?

I'm not sure. Maybe there is an issue in newer Qt where it only binds the first property in such compound statements, thus when reversing them here, it evaluates a binding to `icons` which is having the `iconsChanged` emitted properly, thus causing the whole statement to be evaluated again. But using `onValidChanged` also doesn't work, since it is not being emitted (and perhaps never was being emitted before either).

wdehoog:
So what now? Is it ok like this or do you still want ``valid`` to be removed from the expression? Investigating what is going wrong under the Qt hood is out of my league.

dobey (approved):
Still not sure what exactly is causing the issue, but this seems to resolve it, so let's merge it for now. Really, notifications in UT needs a complete overhaul and most of this logic should be moved into a backend, rather than being part of the shell.

PR #362, opened 2021-01-27
Don't use peeking animation when drawer is open

### LINKED ISSUE #361: Launcher is invisible sometimes
screenshot20210126_221004809

BQ4 5 devel

dobey:
Fixes #361

### DIFF
--- a/qml/Launcher/LauncherPanel.qml
+++ b/qml/Launcher/LauncherPanel.qml
@@ -294,7 +294,7 @@ Rectangle {
                             if(alerting) {
                                 if (!dragging && (launcherListView.peekingIndex === -1 || launcher.visibleWidth > 0)) {
                                     launcherListView.moveToIndex(index)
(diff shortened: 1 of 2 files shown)


Flohack74 (approved):

UniversalSuperBox (changes requested):
The test failure on armhf seems to be a race condition, since it passes on the other architectures:

```
FAIL!  : qmltestrunner::Launcher::test_peekNotRunWhenDrawerOpen() Compared values are not the same
   Actual   (): false
   Expected (): true
   Loc: [/build/unity8-8.20+ubports+0~20210128091109.2~1.gbp5910f1/tests/qmltests/Launcher/tst_Launcher.qml(1641)]
```

dobey:
I've changed the test to use `tryCompare` instead here, and the tests now pass. I also fixed the grammatical faux pas.

UniversalSuperBox (approved):

PR #363, opened 2021-01-28
Fix pulsing notification LED on unread message

UniversalSuperBox:
Prior to this change, hasMessages never became true even when indicator-messages reported a new message.

I'm not entirely sure why this solves the problem.


dobey (approved):
Looks OK.

UniversalSuperBox:
For QA: Does your device's LED blink green when you have a notification? If so, this fix worked.

PR #365, opened 2021-01-29
Fix SurfaceManager assertion errors in qmltests

UniversalSuperBox:
The Shell component is pretty big. No, it's really big. It's got a lot of objects included under it, and it takes a lot of work to create one... or destroy one.

The Shell also happens to have a little component called the SurfaceManager. This object interfaces with QtMir to tell the Shell when applications are started, loaded... the usual.

In the tests, the SurfaceManager is mocked. This mock has a very particular feature: a static `::instance()` property meant to hold a single, global SurfaceManager. Many of the mock's clients, like the mock MirSurface, rely on this static property to simplify their own operation: they don't need to be passed a SurfaceManager at creation time because they can just rely on the global one.

This is not a problem if there is only ever one Shell instance at any time. One Shell instance is created, creating a SurfaceManager. When the Shell instance is destroyed, so is the SurfaceManager. Then you can create a new Shell and SurfaceManager, destroy it, create... until the end of time (or the more likely OOM condition from leaks).

However, Qt 5.12 seems to have changed some timings. This made it so the Loaders used in many of the Lomiri qmltests wouldn't completely destroy their old Shell before the new Shell was loaded. When one Shell loaded before the old one's SurfaceManager was destroyed... boom. The assertion `Q_ASSERT(m_instance == nullptr);` failed and the entire test case aborted.

In Qt 5.9, the [createTemporaryObject](https://doc.qt.io/qt-5/qml-qttest-testcase.html#createTemporaryObject-method) function was added to the TestCase QML type to address exactly this problem. createTemporaryObject ensures that the QML object created is destroyed before the next test begins.

This PR replaces Loaders that load Shells with createTemporaryObject. It simply assigns the temporary object to a TestCase-global `shell` variable. 

I did this in #355 in a slightly more functional way, passing a local Shell object inside the test function. That method seems more correct and would allow us to simplify copied code. It also required a lot more refactoring. This is what I resorted to after I realized it was going to take weeks at the rate I was doing it.

### DIFF
--- a/plugins/GlobalShortcut/globalshortcut.cpp
+++ b/plugins/GlobalShortcut/globalshortcut.cpp
@@ -58,6 +58,9 @@ void GlobalShortcut::setActive(bool active)
 
 void GlobalShortcut::componentComplete()
 {
+    if (this->window()) {
+        QMetaObject::invokeMethod(this, "setupFilterOnWindow", Q_ARG(QQuickWindow*, this->window()));
+    }
     connect(this, &QQuickItem::windowChanged, this, &GlobalShortcut::setupFilterOnWindow);
 }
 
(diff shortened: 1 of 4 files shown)


dobey (approved):
LGTM!

PR #366, opened 2021-01-29
Add 'light' property to IndicatorsModel for tests

UniversalSuperBox:
The qmltests spend a significant amount of time creating the mock indicators on slower hardware. This can take up to 3 seconds per test function on my (2018) laptop, for example. In large test cases like the Shell, it adds up to 35 minutes to complete the test battery...

This PR allows returning less data from the mock indicators model, greatly reducing the object creation time for each test function. The light indicators are used in all tests except the Panel tests, where the Indicators in particular are tested.

This reduces the time to run the test battery from 35 minutes to 23 on my laptop. However, it doesn't seem to have as pronounced an effect on our CI servers, only saving 2 to 3 minutes when the servers are not suffering heavy load. Is that worth the changes to production code? That's what we should discuss...

### DIFF
--- a/plugins/Unity/Indicators/indicatorsmodel.cpp
+++ b/plugins/Unity/Indicators/indicatorsmodel.cpp
@@ -103,6 +103,19 @@ void IndicatorsModel::setProfile(const QString &profile)
     m_manager->setProfile(profile);
 }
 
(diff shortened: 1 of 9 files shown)


dobey (approved):
Looks good!

PR #367, opened 2021-02-03
Changed type of edge-barrier-sensitivity

kugiigi:
Changed type from `y` to `u`

Repeat of #83 
It was merged but apparently, the change was reverted and I don't even see it in the file's history
But whatever, I'm making this again unless it was intended to be reverted :sweat_smile:


dobey (approved):
Looks good. No idea how that got lost. Must have been @mariogrip

PR #315, opened 2020-05-31
NarrowView rotation (no animation)

cibersheep:
Let NarrowView to rotate:

- It uses no animation (yet)
- OSK jumps if rotates when visible in LoginList
- Relies on changes from https://github.com/ubports/unity8/pull/273
- WideView will fail to show Infographics on small screens

screenshot20200531_202924992

image

PR #373, opened 2021-04-15
Enable greeter rotation on landscape devices

kugiigi:
This will enable rotation of the lock screen on devices with landscape as their primary orientation. Rotation is limited on the two landscape orientations since the greeter does not yet have responsive design. This is helpful on tablets since users may use the tablet in either landscape orientations especially when using a magnetic case that can be folded as stand.

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -79,7 +79,7 @@ StyledItem {
 
     readonly property bool orientationChangesEnabled: panel.indicators.fullyClosed
             && stage.orientationChangesEnabled
-            && (!greeter || !greeter.animating)
+            && !greeter.animating
 
     readonly property bool showingGreeter: greeter && greeter.shown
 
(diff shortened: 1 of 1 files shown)


cibersheep:
Oh, I forgot I started this: https://github.com/ubports/unity8/pull/315
Missing: phones, with desktop mode on, fails to show Infographic... we could hide it then

kugiigi:
Yeah, I've seen the other PRs related to greeter rotation but none seem to be ready yet. I was planning to contact the authors of those PR and check if there has been updates or if I can help :)

In any case, this PR is just a quick solution, specially useful for tablets (and for me :D), which can be implemented until full rotation is supported.

cibersheep:
I understand. Ideally would be greeter to rotate on all devices that supports all orientations.

I need to clean, rebase, revise and finish my draft. All help is welcome ^_^

Fuseteam:
is this still needed now #402 has been merged?

kugiigi:
> is this still needed now #402 has been merged?

Well I guess not really needed anymore . Thanks for the heads up

PR #374, opened 2021-04-16
Changed design of side stage handle

kugiigi:
- Used the same handle as the app drawer and indicator panel
- Handle do not overlap the main stage anymore so that it won't cover a few GUs of it.

### DIFF
--- a/qml/Stage/SideStage.qml
+++ b/qml/Stage/SideStage.qml
@@ -26,25 +26,29 @@ Showable {
     property int panelWidth: units.gu(40)
     readonly property alias dragging: hideSideStageDragArea.dragging
     readonly property real progress: width / panelWidth
+    readonly property real handleWidth: units.gu(2)
 
     width: 0
     shown: false
 
-    Item {
+    Handle {
         id: sideStageDragHandle
+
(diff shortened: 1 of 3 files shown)


UniversalSuperBox (approved):
There is still some overlap of the side stage handle's shadow over the main stage app. However, this is much better than before.

cibersheep:
Ouh, this sounds exiting. The current behavior is indeed difficult to hit.
@kugiigi Do you have a screenshot :) pleeeeasseyplease (I love screenshots)

kugiigi:
It's already merged but here's a screenshot anyway 😄
screenshot20210515_005449287

I don't think it's a lot easier to hit though since it's basically the same width. Perhaps it helps a bit visually. There's always a 3-finger tap to toggle side-stage though 😉 

The first version I did, totally removed the shadowy design but I thought I'll keep it to retain the "floating" look of the sidestage in exchange of a negligible overlap 😄

cibersheep:
You are very kind <3
It looks more clear than the shadow to be honest.

PS: You can also drag an app with 3 finger to the stage ^__^

PR #375, opened 2021-04-16
Corrected terminal app ID

kugiigi:
This will make `Ctrl + Alt + T` actually work with UT's terminal app.


UniversalSuperBox (approved):
Yep, that launches the terminal.

PR #377, opened 2021-04-16
Hide hint when closing and not empty

kugiigi:
This hides the side-stage hint when an app is active and when closing animation is running.


UniversalSuperBox (changes requested):
Please:

1. File an issue report giving steps how to reproduce this problem
2. Find a way to reproduce the problem in Lomiri's test suite. The test should fail without your changes.
3. Apply your changes and ensure they fix the problem as stated. Commit the test and the changes in the same commit, or add the test in a second commit.

PR #378, opened 2021-04-19
Always hide launcher in lock screen

kugiigi:
This hides the launcher in lock screen even with desktop mode enabled and auto-hide disabled.

### DIFF
--- a/qml/Shell.qml
+++ b/qml/Shell.qml
@@ -583,7 +583,7 @@ StyledItem {
             superPressed: physicalKeysMapper.superPressed
             superTabPressed: physicalKeysMapper.superTabPressed
             panelWidth: units.gu(settings.launcherWidth)
-            lockedVisible: (lockedByUser || shell.atDesktop) && lockAllowed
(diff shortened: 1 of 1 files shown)


cibersheep:
What's the idea behind this?

Is not `System Settings > Lock & Unlock > Launcher -> off` enough?

kugiigi:
No, when in desktop mode and launcher auto-hide is disabled, the launcher is shown even in lock screen.
Regardless of the value set in `System Settings > Lock & Unlock > Launcher`.
And in my opinion, the launcher should be hidden in lock screen anyway.

cibersheep:
Ah, I didn't understand it. It's clear now. Thanks

UniversalSuperBox (changes requested):
Please:

1. File an issue report giving steps how to reproduce this problem
2. Find a way to reproduce the problem in Lomiri's test suite. The test should fail without your changes.
3. Apply your changes and ensure they fix the problem as stated. Commit the test and the changes in the same commit, or add the test in a second commit.

PR #379, opened 2021-04-19
Disable shortcuts when launcher is not available

kugiigi:
This disables keyboard shortcuts when the launcher is disabled or not available.

i.e. Launcher is disabled in lock screen.
Currently, swiping won't show the launcher if it's disabled in lock screen but keyboard shortcuts are still working.

### DIFF
--- a/qml/Launcher/Launcher.qml
+++ b/qml/Launcher/Launcher.qml
@@ -72,7 +72,7 @@ FocusScope {
         if (state == "drawer")
             return;
 
-        if (superPressed) {
+        if (superPressed && root.available) {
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
This PR looks good to me, but I believe a unit test should be added to ensure that the shortcuts do not function when the shell is locked. This will ensure that we don't add this bug back in the future.

dobey (changes requested):
Please add a test to ensure this doesn't break in the future, as Dalton requested.
Also, it might be good to clean up the commits by squashing with `git rebase -i` after.

PR #383, opened 2021-05-03
Fixed #251

### LINKED ISSUE #251: 4 finger tap summons both the osk and the drawer
could be related to #22 :p 
but tapping with 4 finger opens the drawer and while that happens the osk pops up before being dismissed

kugiigi:


UniversalSuperBox:
I don't think this is correct. I developed the automatic focus of the search field so someone could press Super on a physical keyboard, type what they want, then press Enter to submit. If the search field is gaining focus on a four-finger press, I think the thing that responds to that gesture should be fixed so that the search field doesn't get focused in the first place.

You will also want to fix `test_fourFingerTapOpensDrawer` so it tests that the search field is not focused on a four-finger tap.

kugiigi:
The code [here](https://github.com/kugiigi/unity8/blob/b058db85441929c049a62808c73396735f4b979e/qml/Shell.qml#L370) passes `true` so my impression is that it's the desired behavior and by design. And personally, I like it that way because I like searching instead of looking for an icon in a sea of icons :grinning:

UniversalSuperBox:
Hmm, I don't think so. I think that a four-finger tap should be the same as a Super key press or a tap/click on the bfb. The keyboard appearing on this press seems surprising to me. While it *might* be by design, it might have also been the original developer filling in the gaps.

cibersheep:
Mhm. Difficult question here.

On Unity7 when you trigger the Dash (bfb, super key) it focuses the search automatically. So I got the sense that Lomiri's Dash worked the same (sometimes the focus is persists and the OSK get triggered).

This behavior is good in big screens but not so much on the phone where about half of the screen is covered with the OSK. This might be a discussion of «another flour bag» though.

The main thing here is: long swipe, tap on bfb, and on home (Arale's home button that is kind of a super key) shows the Dash with no focus on the Search area.
I think 4-fingers tap should stick to the same behavior and work in the same manner: showing the Dash

UniversalSuperBox:
Also note that, with a hardware keyboard, typing any letter will immediately focus the search box and begin typing. The box just isn't focused so the on-screen keyboard doesn't appear by default.

kugiigi:
Okay, I guess I'm in the minority here 😅
To be fair, 4-finger tap would only be used on large screens so OSK covering large portion of the screen isn't too bad. But, anyway it is what it is. I'll just throw a tantrum and not do the accepted fix since it's not to my liking 😂

It'll be simple though, just pass `false` to the line I linked then the test if still needed.

UniversalSuperBox:
Superseded by https://github.com/ubports/unity8/pull/391. I accidentally fixed the race condition that resulted in the keyboard appearing then disappearing again.

PR #390, opened 2021-05-31
PhysicalKeysMapper: handle first power button long press correctly

peat-psuwit:
When a power button pressing is the very first input to Lomiri/Unity8,
the PhysicalKeysMapper won't recognize the long pressing correctly.

Turns out, the timestamp Qt gives us can start at 0, so using 0 as an
invalid value breaks this case. So, use -1 instead.

The test is updated to also test this case.

### DIFF
--- a/qml/Components/PhysicalKeysMapper.qml
+++ b/qml/Components/PhysicalKeysMapper.qml
@@ -65,7 +65,7 @@ Item {
         property bool superPressed: false
         property bool superTabPressed: false
 
-        property var powerButtonPressStart: 0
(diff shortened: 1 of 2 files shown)


peat-psuwit:
For QA:
1. Restart Unity8. If you don't know how, just restart the phone. Make sure you don't touch or press anything during the startup for reproducibility.
2. When the shell is fully up, long-press the power button, as if you're going to turn off the phone.
3. After ~2 seconds, the power dialog should appear.

UniversalSuperBox (approved):
Does what it says on the tin.

PR #391, opened 2021-06-01
Fix edge cases in Drawer

### LINKED ISSUE #251: 4 finger tap summons both the osk and the drawer

### LINKED ISSUE #389: Drawer search contents seem to always miss the first letter typed on a physical keyboard
A common desktop usage method for power users and enthusiasts goes a bit like this:

"I want to launch Firefox"
1. Press the Super key
2. Type "fire"
3. Press Enter

However, there is often a problem when trying this with a physical keyboard on Lomiri: Instead of "fire" being listed in the search box at the end of step 2, it only says "ire" or "re"!

This problem is triggered when the user is running too fast for Lomiri. When the Super key is pressed, an animation is started on the Drawer. During that animation, it's already possible to type in the Drawer search box. Once the animation completes, the contents of the Drawer's search box are removed. This behavior was introduced in https://github.com/ubports/unity8/commit/c1712df8f7fd0b3c5aad210811cc52907dc2ba5d

UniversalSuperBox:
This started as fixing #389, a strange edge case where you could type *too fast* for the Drawer during its opening animation. It turned into fixing several more edge cases.

Trying to write a test case that canceled the Drawer drag closed resulted in me finding c6628ec. To make sure the Drawer was _really_ opened all the way, I dragged it way past its fullyOpen boundary. It took me quite some time to figure out why the tests couldn't close it again!

Fixing #389 revealed a case where the Drawer's focus being removed and set in quick succession would cause the search box to lose its selection, resulting in the text cursor locating itself after the final selected character instead. This most commonly happened when starting to drag the Drawer closed, then changing your mind and dragging it all the way back open. Normally, fullyOpen becomes true long after draggingHorizontally becomes false due to the animation. This means that the Launcher's animateTimer fires before fullyOpen becomes true. The fully open case resulted in fullyOpen and draggingHorizontally changing immediately, with the Launcher's animateTimer firing on the next frame. This ordering switch resulted in the Drawer's search box getting its focus and cursor set correctly right before the entire Drawer lost focus, resetting the search box.

Also somehow this fixed https://github.com/ubports/unity8/issues/251. I didn't exactly like this behavior on small screens, but hey it is what it is.

Fixes https://github.com/ubports/unity8/issues/389
(Accidentally) Fixes https://github.com/ubports/unity8/issues/251

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2016 Canonical, Ltd.
- * Copyright (C) 2020 UBports Foundation.
+ * Copyright (C) 2020-2021 UBports Foundation
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


dobey (approved):
Looks good. ++ for the tests.

PR #394, opened 2021-06-05
Reduce ram usage by 25%, improve startup performance, make Greeter prettier.

UniversalSuperBox:
This PR resolves the single greatest user of RAM by percentage, the wallpaper. It also resolves the single heaviest cause of startup lag by percentage, the thing that checks all of the user's desired wallpapers to make sure they actually load. And it resolves the position of the Greeter wallpaper.

That's right, three types of resolution.

Resolving the Greeter wallpaper means positioning it correctly. Have you ever noticed that the wallpaper jiggles a bit when you unlock your device? That's because the Greeter's wallpaper image is actually shorter than the one that serves as your background image. The first commit here makes sure the Greeter's image and the one that serves as the Stage's background match. I took extra care to make sure that the clock and InfoGraphic didn't move on a narrow (phone) greeter. The InfoGraphic and login box moved a bit on the wide greeter... they're actually in the vertical center of the screen now.

Next we optimize the biggest RAM hog in Lomiri by setting sourceSize on its wallpapers to an appropriate value. You can get a whole blog post about why that's needed and why it's done the way it is in the second commit's message. But for the stats: This saves 60MB of RAM when Lomiri is run at 540x960 (the resolution of the BQ E4.5) with software rendering and with the default wallpaper. That was ~25% of its memory usage at that resolution. The savings go up if your wallpaper is larger than the default wallpaper's beefy 4096x2304. They go down if your device has a higher resolution than 540x960 or if your wallpaper is smaller than the default. But in all cases, *it's free RAM*. This commit required special care to make sure that all images were loaded at the same resolution. Not doing so would incur multiple image loads from disk and squander our saved RAM.

The final commit starts by reducing Shell creation time by about 10%. How that translates into real-world startup performance I'm not sure, but it saves about a minute in the tests on CI. We do that by first optimizing WallpaperResolver. It loaded the final image in its candidates to make sure it could, but it returned the final image whether or not the load was successful. So there was no reason to load it in the first place. Then, we make all of the Images between the Wallpaper and the WallpaperResolver asynchronous. This means that the shell no longer freezes when it's changing wallpapers, it starts up just a little bit faster, and the Greeter appears sooner when requested. It had an unfortunate side effect, though: the wallpaper would be transparent for a moment and then pop in. I fixed that by adding a small fade-in animation that plays after the image loads. When I showed it to people, they asked me "what changed?" so I think the UX is good enough.

The final commit could probably be split into 3, but I don't think there's any reason to. They would conflict with each other on revert anyway.


qml/Components/Wallpaper.qml
```diff
@@ -20,12 +21,76 @@ import Ubuntu.Components 1.3
...
     Image {
+        id: image
+        objectName: "wallpaperImage"
         anchors.fill: parent
         fillMode: Image.PreserveAspectCrop
 
-        source: root.source
+        asynchronous: true
+
+        property real oldSourceSize
```
mardy: I'm not sure whether it's needed, but just to be on the safe side I'd initialize this to `0.0`

qml/Components/WallpaperResolver.qml
```diff
@@ -23,23 +24,35 @@ Item {
     // Provide a list of wallpapers to resolve here, preferred ones first
     property var candidates: []
 
+    property bool cache: true
+
     readonly property url background: {
         for (var i = 0; i < repeater.count; i++) {
-            if (repeater.itemAt(i).status === Image.Ready)
+            var image = repeater.itemAt(i);
+            if (image === null)
```
mardy: does this actually happen? It seems an impossible case.

qml/Components/Wallpaper.qml
```diff
@@ -20,12 +21,76 @@ import Ubuntu.Components 1.3
...
+        color: theme.palette.normal.background
+        anchors.fill: parent
+        opacity: image.status === Image.Ready ? 0: 1
+        visible: opacity !== 0
+        Behavior on opacity {
+            UbuntuNumberAnimation { duration: UbuntuAnimation.FastDuration }
+        }
+    }
+
+    Timer {
```
mardy, line 86: If you used a delayed binding for the source property, you could get rid of this.

mardy (approved):
Thanks for the tests!!

I added a few comments inline, but nothing that should prevent this from being merged.

qml/Components/Wallpaper.qml
UniversalSuperBox: I'm not sure I understand. What am I replacing with a delayed binding?

If you mean the timer, it's there because the Wallpaper will just *not load* if its sourceSize is unset. I know that it'll bite me or someone else working on Lomiri in the future (because it's such a dumb non-obvious effect) and so I added the warning.

qml/Components/WallpaperResolver.qml
UniversalSuperBox: This may have been a relic from when I was trying to use `root.candidates.slice(0, -1)` instead of `repeater.count` for this for loop. That actually caused some unexpected behavior in QML, the binding for `background` stopped firing when objects inside the repeater changed. I went back to `repeater.count` and it fixed the issue.

I'll fix it up.

qml/Components/Wallpaper.qml
```diff
@@ -20,12 +21,76 @@ import Ubuntu.Components 1.3
...
+        }
+
+        onIntermediarySourceSizeChanged: {
+            if (oldSourceSize !== 0.0 && oldSourceSize !== -1.0 && root.source !== "") {
+                console.warn(`${root.objectName} source size is changing from ${oldSourceSize} to ${intermediarySourceSize}. This will cause an expensive image reload.`);
+            }
+            oldSourceSize = intermediarySourceSize;
+        }
+
+        sourceSize: Qt.size(intermediarySourceSize, intermediarySourceSize)
```
UniversalSuperBox, line 70: I've been thinking about this more. Since the wallpaper loading is now asynchronous, we could save more memory by passing sourceSize as a `size` based on the current orientation of the shell. The image would have to reload every time the user rotated their phone, but it wouldn't cause a UI thread pause now that the images are async.

I haven't benchmarked it to see how much memory we'd save, but in theory it could be worth it. Maybe I should wait until after we merge this though? It'd be pretty easy to do, just replacing `real sourceSize` / `real backgroundSourceSize` with `size` values instead.

UniversalSuperBox: I'll try it out later. This is a large enough RAM savings to warrant doing first without waiting for more development.

mardy: I wouldn't even try it. I'm not convinced that you'll get any savings there, and you risk making the rotation animation slower. I think this is very good the way it is :-)

qml/Components/Wallpaper.qml
mardy: I meant that instead of using a Timer here, you could have the `source` changed signal connected with a delayed binding, and at that point in the handler of this binding you could be sure that the sourceSize must have been set. If it isn't set, then you'd print the error message. But it's a minor thing, so do it only if you like: I'm approving this anyways.

mardy (approved):

lduboeuf:
humm, when having the default wallpaper, i see the screen turn grey-ish just before PIN code screen

UniversalSuperBox:
When? Could you provide a screenshot and steps to reproduce?

lduboeuf:
hard to take a screen shot, but after boot and after the "enter pin code" notification and just before the PIN code screen , the screen goes grey for 1 second.
I have that kind of background which is the default one if i'm not wrong:
screenshot20210609_150054732

UniversalSuperBox:
Yes, this is the same as all Wallpapers. When they need to reload, they start as gray then fade in once the image is ready. That means that the first Wallpaper you see in Lomiri (be it the PIN unlock or the cover page) will be gray for a moment then fade in.

PR #398, opened 2021-06-16
Add Clickable support and update readme

UniversalSuperBox:
No, this doesn't mean that you can actually build a click of Lomiri and install it on your device just like *that*. Nor can you use `clickable desktop` to act like a little Ubuntu Touch emulator. Those would be cool. They just aren't what this does.

This PR adds configuration that allows Clickable to build and run the tests for Lomiri. This greatly reduces the number of hoops that a contributor needs to jump through to start testing their changes to Lomiri. Where most contributors would just throw out the PR and hope before, now they can run the tests and try their changes locally.

Thanks to Jonny for making this possible. He was able to snap together the pieces that Clickable provides to make a completed puzzle.


README.md
```diff
@@ -1,51 +1,31 @@
...
-
-
-### Get involved
-
-* [Kanban-Board](https://github.com/ubports/unity8/projects/1)
-* If you are interesting in helping us in any way you can join [Unity8 Forum](https://forums.ubports.com/category/36/unity8).
-* For reporting bugs or filing feature-requests, please use the project’s [issues page](https://github.com/ubports/unity8/issues).
-* For IM you can join our Telegram group [for Developer users](https://t.me/UBports_Unity8).
+* `clickable ide qtcreator` will open QtCreator with this repository open as a project. You can edit and build the project this way, but running the tests will be a bit difficult.
+* `clickable build-libs --dirty --arch=amd64` will build Lomiri in the same environment as it would receive in Ubuntu Touch.
```
jonnius: No need for `--arch=amd64` because the architecture is restricted by the clickable.json `restrict_arch` field. The `--dirty` flag can be removed as well, because `build-libs` does not include cleaning the build dir.

clickable.json
```diff
@@ -0,0 +1,158 @@
+{
+    "builder": "custom",
+    "build": "echo Run \\'clickable build-libs --debug\\' to build && false",
```
jonnius, line 3: The instructions above do not include the `--debug` flag. If it is not needed, it should be removed in this help message as well.

clickable.json
```diff
@@ -0,0 +1,158 @@
...
+    "restrict_arch": "amd64",
+    "libraries": {
+      "unity8": {
+        "builder": "cmake",
+        "src_dir": "${ROOT}",
+        "test": "make xvfballtests -j${NUM_JOBS}",
```
jonnius: Where does `NUM_JOBS` come from? I guess this will run with unlimited build jobs. But you can use [`NUM_PROCS`](https://clickable-ut.dev/en/latest/clickable-json.html#placeholders-environment-variables) instead.

UniversalSuperBox: My mistake, I fumbled the name of the variable.

clickable.json
UniversalSuperBox: True, --debug should be on so debugging symbols and the QML debugger are enabled.

README.md
UniversalSuperBox: Got it.

jonnius:
Looks good to me now. :)

Issue #397, opened 2021-06-16
Scrolling through users in desktop mode doesn't change login method


capsia37:
Hello!
When switching to a different user using mouse scroll instead of clicking on the name of the user the login screen changes the username, but it doesn't update the login field(s).
![Screenshot](https://user-images.githubusercontent.com/43005909/122278906-bd157780-cee7-11eb-9b00-c8cd56a301ca.png)

PR #401, opened 2021-07-02
Greeter redesign

capsia37:
This pull request contains new styles for the greeter prompts, the changes include:

Separation of prompts from GreeterPrompt.qml to multiple files:
- PinPrompt.qml : a new prompt for pin showing empty circles that are filled when typing
- TextPrompt.qml : a redesigned prompt for generic text input, usually a password field
- ButtonPrompt.qml : a prompt that the user only needs to press to login. This prompt is only used in multi user mode. Single user devices will autologin when no password is set

Changes and fixes to related issues:
- Moving the cursor and selecting the text is now disabled in the password / pin field
- Fixes for keyboard detection: the keyboard icon won't show on the prompts if no keyboard is detected
- Fix keyboard tests
- Addition of a user icon near the username
- Replace white background with transparency. Blur and darkening is added to prevent unreadable text
- Move session selector to the bottom right to free up space
- Fix scrolling on login list #397
- Adjust colors to make everything visible

### DIFF
--- a/qml/Greeter/ButtonPrompt.qml
+++ b/qml/Greeter/ButtonPrompt.qml
@@ -5,23 +5,21 @@ import "../Components"
 FocusScope {
     id: root
     objectName: "promptButton"
-    focus: true
 
-    property string text
+    property alias text: buttonLabel.text
+    property alias interactive: root.enabled
     property bool isSecret
-    property bool interactive: true
     property bool loginError: false
     property bool hasKeyboard: false
     property string enteredText: ""
-    property bool inputFocus
 
     signal clicked()
     signal canceled()
     signal accepted(string response)
 
-    Keys.onSpacePressed: triggered();
-    Keys.onReturnPressed: triggered();
-    Keys.onEnterPressed: triggered();
+    Keys.onSpacePressed: clicked();
+    Keys.onReturnPressed: clicked();
+    Keys.onEnterPressed: clicked();
 
     anchors.fill: parent
 
@@ -35,25 +33,19 @@ FocusScope {
             ColorAnimation{}
         }
         border {
-            color: loginError ? theme.palette.normal.negative : theme.palette.normal.raisedSecondaryText
-            width: loginError ? units.dp(3): units.dp(2)
-        }
-    }
-
-    function triggered() {
-        if (root.interactive) {
-            root.clicked();
+            color: root.loginError ? theme.palette.normal.negative : theme.palette.normal.raisedSecondaryText
+            width: root.loginError ? units.dp(3): units.dp(2)
         }
     }
 
     MouseArea {
         anchors.fill: parent
-        onClicked: parent.triggered();
+        onClicked: parent.clicked();
     }
 
     Label {
+        id: buttonLabel
         anchors.centerIn: parent
         color: theme.palette.normal.raisedSecondaryText
-        text: root.text
     }
 }
--- a/qml/Greeter/ButtonPrompt.qml
+++ b/qml/Greeter/ButtonPrompt.qml
@@ -1,0 +1,58 @@
+import QtQuick 2.4
+import Ubuntu.Components 1.3
+import "../Components"
+
+FocusScope {
+    id: root
+    objectName: "promptButton"
+    focus: true
+
+    property string text
+    property bool isSecret
+    property bool interactive: true
+    property bool loginError: false
+    property string enteredText: ""
+    property bool inputFocus
+
+    signal clicked()
+    signal canceled()
+    signal accepted()
+
+    Keys.onSpacePressed: triggered();
+    Keys.onReturnPressed: triggered();
+    Keys.onEnterPressed: triggered();
+
+    anchors.fill: parent
+
+    activeFocusOnTab: true
+
+    Rectangle {
+        anchors.fill: parent
+        radius: units.gu(0.5)
+        color: "#7A111111"
+        Behavior on border.color {
+            ColorAnimation{}
+        }
+        border {
+            color: loginError ? theme.palette.normal.negative : theme.palette.normal.raisedSecondaryText
+            width: loginError ? units.dp(3): units.dp(2)
+        }
+    }
+
+    function triggered() {
+        if (root.interactive) {
+            root.clicked();
+        }
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        onClicked: parent.triggered();
+    }
+
+    Label {
+        anchors.centerIn: parent
+        color: theme.palette.normal.raisedSecondaryText
+        text: root.text
+    }
+}
--- a/qml/Greeter/ButtonPrompt.qml
+++ b/qml/Greeter/ButtonPrompt.qml
@@ -11,6 +11,7 @@ FocusScope {
     property bool isSecret
     property bool interactive: true
     property bool loginError: false
+    property bool hasKeyboard: false
     property string enteredText: ""
     property bool inputFocus
 
--- a/qml/Greeter/ButtonPrompt.qml
+++ b/qml/Greeter/ButtonPrompt.qml
@@ -34,7 +34,7 @@ FocusScope {
         }
         border {
             color: root.loginError ? theme.palette.normal.negative : theme.palette.normal.raisedSecondaryText
-            width: root.loginError ? units.dp(3): units.dp(2)
+            width: root.loginError ? units.dp(2): units.dp(1)
         }
     }
 
--- a/qml/Greeter/ButtonPrompt.qml
+++ b/qml/Greeter/ButtonPrompt.qml
@@ -17,7 +17,7 @@ FocusScope {
 
     signal clicked()
     signal canceled()
-    signal accepted()
+    signal accepted(string response)
 
     Keys.onSpacePressed: triggered();
     Keys.onReturnPressed: triggered();
--- a/qml/Greeter/CoverPage.qml
+++ b/qml/Greeter/CoverPage.qml
@@ -119,7 +119,7 @@ Showable {
         anchors.fill: effectSource
         source: effectSource
         radius: 64
-        transparentBorder: true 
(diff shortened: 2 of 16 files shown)


lduboeuf:
awesome, can you provide some sample screen shots ?

capsia37:
> awesome, can you provide some sample screen shots ?

sure, here you are :)

Password login on mobile:
Schermata da 2021-07-02 13-19-54

With the password:
Schermata da 2021-07-02 13-20-09

Pin login:
Schermata da 2021-07-02 13-20-44

Failed pin login:
Schermata da 2021-07-02 13-20-58

Button login on desktop:
Schermata da 2021-07-02 13-22-41

Background blur ( pin on desktop)
Schermata da 2021-07-02 13-22-28

cibersheep:
Oh. That's nice! Happy to see the changes :)
A couple of small things:
-Isn't the border for text areas and login button too thick?
-«Has XXX» text on phone view looks a bit too big as well (in desktop mode looks more compensated

The rest is looking very good

capsia37:
Hello, and thank you for the feedback!

I agree with both of the points, I've noticed that the sizes are rendered a bit different on device than on the Lomiri test shell, so on device the borders actually already look a bit thinner. 

On Nexus 5:
screenshot20210703_200326432

I'm using bold username text because mobile devices won't have background blur to reduce memory and computational power usage and the simplest way to keep the username visible also on different backgrounds was to make it a bit bigger and bolder. Any better idea is appreciated :)

cibersheep:
Gimme a day and I reply properly :) (does that sound scary? Hi hihih)

qml/Greeter/ButtonPrompt.qml
```diff
@@ -0,0 +1,59 @@
...
+    objectName: "promptButton"
+    focus: true
+
+    property string text
+    property bool isSecret
+    property bool interactive: true
+    property bool loginError: false
+    property bool hasKeyboard: false
+    property string enteredText: ""
+    property bool inputFocus
```
mardy: This seems unused

qml/Greeter/ButtonPrompt.qml
```diff
@@ -0,0 +1,59 @@
...
+import Ubuntu.Components 1.3
+import "../Components"
+
+FocusScope {
+    id: root
+    objectName: "promptButton"
+    focus: true
+
+    property string text
+    property bool isSecret
```
mardy, line 11: this seems unused

qml/Greeter/ButtonPrompt.qml
```diff
@@ -0,0 +1,59 @@
+import QtQuick 2.4
+import Ubuntu.Components 1.3
+import "../Components"
+
+FocusScope {
+    id: root
+    objectName: "promptButton"
+    focus: true
+
+    property string text
```
mardy: this could be an alias to the label's text

qml/Greeter/ButtonPrompt.qml
```diff
@@ -0,0 +1,59 @@
...
+    Rectangle {
+        anchors.fill: parent
+        radius: units.gu(0.5)
+        color: "#7A111111"
+        Behavior on border.color {
+            ColorAnimation{}
+        }
+        border {
+            color: loginError ? theme.palette.normal.negative : theme.palette.normal.raisedSecondaryText
+            width: loginError ? units.dp(3): units.dp(2)
```
mardy: here and in the line above, it's better to be explicit and replace `loginError` with `root.loginError`

qml/Greeter/ButtonPrompt.qml
```diff
@@ -0,0 +1,59 @@
...
+import "../Components"
+
+FocusScope {
+    id: root
+    objectName: "promptButton"
+    focus: true
+
+    property string text
+    property bool isSecret
+    property bool interactive: true
```
mardy: this can be an alias to the `enabled` property. Then you can remove the `triggered()` function and in the MouseArea do just:
```
    MouseArea {
        anchors.fill: parent
        onClicked: root.clicked();
    }
```

cibersheep:
I think I would go with something in this direction a bit more subtle.
About the background color problem, I guess you discarted the lightdm grey box :) so that could let us two routes:
-coloring the text depending on the background (there's a function for that in the uuitk (which is used for changing the color of the text of buttons accordingly to its background color)
-diming the background (on the mockup, it has a 10-20% more attenuation)

imatge

qml/Greeter/ButtonPrompt.qml
capsia37: Yes, it is unused, but not adding this will cause some warnings to show, this is because this file is loaded by the same loader as PinPrompt.qml and TextPrompt.qml. I could add a 'when' on the binding, but this will make the code more complex if we decide to add new login methods ( pattern, biometrical recognition... ).

qml/Greeter/ButtonPrompt.qml
capsia37: Hello and thank you for the suggestions. 
The function triggered() is also called when enter/space is pressed on the keyboard, what should I do with the keybindings?

qml/Greeter/ButtonPrompt.qml
mardy: I see, but it seems to me that this variable in all login methods is either unused or not necessary: if the parent item is a FocusScope and the Input field has `focus: true` set, then the field will get the focus everytime the parent `root` element has the focus. So, instead of having an `inputFocus` property on the root element, you can just play with the root element's `focus` property, and if you need to explicitly move focus to it, you can just call `forceActiveFocus` on the root element.

qml/Greeter/ButtonPrompt.qml
mardy: you can just directly call `root.clicked()`, because the key presses will not be received if the item is disabled.

qml/Greeter/ButtonPrompt.qml
capsia37: Ok, thank you! The only thing I still need is the isSecret property for the TextPrompt. It determines whether the input should be plain text or hidden. What to you suggest for that?

capsia37:
Hello @cibersheep sorry for the late response, what do you think?
screenshot

@mardy I've tried to add all the suggestions you've made but removing the focus variable seems to break focus and the stage will get focus instead of the greeter. The isSecret variable is needed in the TextPrompt in case we will have different auth methods.

cibersheep:
> Hello @cibersheep sorry for the late response, what do you think?

No worries! That looks really good. Thanks for your work!

UniversalSuperBox:
This PR now conflicts on the `.gitignore` and `clickable.json` added by https://github.com/ubports/unity8/pull/398. Please update this PR to drop d8c48c4.

capsia37:
> This PR now conflicts on the `.gitignore` and `clickable.json` added by #398. Please update this PR to drop [d8c48c4](https://github.com/ubports/unity8/commit/d8c48c4bc4ff88e81483154f62564473c4c6551a).

@UniversalSuperBox done in all PRs

Flohack74:
Will this be ready for OTA-20 ?

capsia37:
> Will this be ready for OTA-20 ?

@Flohack74 That would be great! I'm not sure about how to solve the last two suggestions that @mardy wrote in the review.

Flohack74:
Alright I target it now for 20 ^^

Flohack74:
@mardy can you have a look at your last 2 suggestions and help out Capsia a bit? Thx!

mardy:
> @mardy can you have a look at your last 2 suggestions and help out Capsia a bit? Thx!

I don't remember what they were :-) But if I remember correctly they were just minor suggestions for the code, nothing that should block this from landing.

What I'm concerned about is that this PR is huge, and also the commit history is not clean. I wounder if it would be possible to create a few smaller PRs instead; not only reviewwing will be easier, but also reverting (in case some parts of this code turn out to cause regressions).

Flohack74:
Humm, yes, huge feature branches should be avoided :P - I have to give a lesson about trunk-based development probably. Also feature flags can save the day, you can review and commit in small steps, and things are already merged "under the hood" but not activated until golive.

That said it will be hard to drag this apart. As we are at the beginning of the next OTA cycle we can also break devel if it breaks, no problem, thats what we always said ;)
@capsia37 one thing, on the phone would the username being shown always be "phablet" ?

capsia37:
@Flohack74 
> @capsia37 one thing, on the phone would the username being shown always be "phablet" ?

On the phone ( and also when the greeter is set in single user mode ) the username is set to the device name. This is the same behavior as before.

I think I can separate this in multiple PRs if needed :)

Flohack74:
Nah lets merge it and see what happens ;)

Flohack74 (approved):

ivoxavier:
The update went well, so far is looking go. Tested with pin code and passphrase no issues identified.

There's only the discussion about if has or not the rotation feature.

Tested on FP2: Devel channel image (2021-11-24)

cibersheep:
It seems to work great.

One ¿issue? is when you get pin wrong for too many times, it will make you wait for five minutes (so far so good).
 
After that period, trying to unlock the phone again, the alert icon will be shown beside the «pin circles». Is that expected?

Flohack74:
@ivoxavier indeed rotation is in a separate PR and we try to merge it too ;)

capsia37:
> It seems to work great.
> 
> One ¿issue? is when you get pin wrong for too many times, it will make you wait for five minutes (so far so good).
> 
> After that period, trying to unlock the phone again, the alert icon will be shown beside the «pin circles». Is that expected?

I think this was the previous behavior, but it should be fixed

c4pp4:
I think circles should be filled with white color after whole pin input, now it's weird:

Flohack74:
@c4pp4 I agree thats a bit of a UI glitch, can you plz open a new issue? Thx!

PR #403, opened 2021-07-07
Notch support

capsia37:
This PR contains code changes to support devices with a notch without the need of tweaks:
- Added top margin on the shell to prevent the notch from overlapping the indicators
- Read parameter for top margin height from the device configuration
- Added a Gsetting to allow reverting changes to the previous behavior

This PR is ready as a fix for the issue, but integrating the notch permanently in UT would need some more discussion and design.


UniversalSuperBox:
This PR now conflicts on the `.gitignore` and `clickable.json` added by https://github.com/ubports/unity8/pull/398. Please update this PR to drop d8c48c4.

PR #404, opened 2021-07-14
Add shortcuts for side stage

### LINKED ISSUE #395: Keyboard shortcuts for side-stage
### Description
I think side-stage should also have keyboard shortcuts (opening, closing, moving apps in and out, etc.)

### Justification
Convergent design :)
And there are times especially on tablets where side-stage setup is better than windowed mode.

kugiigi:
- Added `Super + S` to show/hide the side stage.
Not sure how acceptable that shortcut is. I used it since it's called the side stage :grinning: 
- Repurposed `Ctrl + Super + Left/Right` for moving an app between the main stage and side stage
- Added `Stages` section in the shortcuts overlay for these new shortcuts

This fixes #395

### DIFF
--- a/qml/Components/KeyboardShortcutsOverlay.qml
+++ b/qml/Components/KeyboardShortcutsOverlay.qml
@@ -241,19 +241,6 @@ Rectangle {
                 }
 
                 Label {
-                    text: i18n.tr("Super + S")
-                    fontSize: "small"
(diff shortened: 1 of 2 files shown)


UniversalSuperBox (changes requested):
Please add tests which send these inputs to the desired component and ensure the results are as you expect.

PR #411, opened 2021-08-19
Change screenshotter sound effect path to our own sound package

Flohack74:
The screenshotter was using the shutter sound from inside the Android container. This broke recently with Halium 9 devices, but also on Pinephone this would never have worked. Its sane to expect those files are shipped by the rootfs, not a Halium container.
Note that the shutter sound as well as the focus sound which is not currently used are different from what we have now out of AOSP, but after so many years with that one sound I think we can make a change and celebrate that it will work now on any device correctly.

### DIFF
--- a/qml/Components/ItemGrabber.qml
+++ b/qml/Components/ItemGrabber.qml
@@ -35,7 +35,7 @@ Rectangle {
 
     NotificationAudio {
         id: shutterSound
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
Depends on https://gitlab.com/ubports/core/ubuntu-touch-sounds/-/merge_requests/3 being merged

dobey (approved):
Looks OK.

PR #414, opened 2021-10-02
Test and use hfd if Android SDK indicates it

Flohack74:
To get LEDs working again after H9 and most newer devices requiring that API. See also https://gitlab.com/ubports/community-ports/android10/project-management/-/issues/6

### DIFF
--- a/plugins/Lights/HfdLights.cpp
+++ b/plugins/Lights/HfdLights.cpp
@@ -18,14 +18,13 @@
 
 #include "HfdLights.h"
 #include <QtCore/QDebug>
-#include <QProcess>
 #include <QDBusArgument>
 #include <QDBusConnection>
 #include <iostream>
 
-extern "C" {
-#include <string.h>
-}
+#define HFD_SERVICE_NAME "com.lomiri.hfd"
+#define HFD_SERVICE_PATH "/com/lomiri/hfd"
+#define HFD_SERVICE_INTERFACE "com.lomiri.hfd.Leds"
 
 HfdLights::HfdLights(QObject* parent)
   : Lights(parent)
--- a/plugins/Lights/HfdLights.cpp
+++ b/plugins/Lights/HfdLights.cpp
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2021 UBports Foundation.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Florian Leeber <florian@ubports.com>
+ */
+
+#include "HfdLights.h"
+#include <QtCore/QDebug>
+#include <QProcess>
+#include <QDBusArgument>
+#include <QDBusConnection>
+#include <iostream>
+
+extern "C" {
+#include <string.h>
+}
+
+HfdLights::HfdLights(QObject* parent)
+  : Lights(parent)
+{
+        m_hfdInterface = new QDBusInterface(
+            QStringLiteral(HFD_SERVICE_NAME),
+            QStringLiteral(HFD_SERVICE_PATH),
+            QStringLiteral(HFD_SERVICE_INTERFACE),
+            QDBusConnection::systemBus(),
+            this
+        );
+}
+
+bool HfdLights::init()
+{
+    // hfd does not need any initalization to be used
+    return true;
+}
+
+void HfdLights::turnOn()
+{
+    m_hfdInterface->call(QStringLiteral("setState"), 0);
+    m_hfdInterface->call(QStringLiteral("setColor"), (quint32)m_color.rgba());
+    m_hfdInterface->call(QStringLiteral("setOnMs"), m_onMs);
+    m_hfdInterface->call(QStringLiteral("setOffMs"), m_offMs);
+    m_hfdInterface->call(QStringLiteral("setState"), 1);
+}
+
+void HfdLights::turnOff()
+{
+    m_hfdInterface->call(QStringLiteral("setState"), 0);
+    m_hfdInterface->call(QStringLiteral("setColor"), (quint32)0);
+    m_hfdInterface->call(QStringLiteral("setOnMs"), 0);
+    m_hfdInterface->call(QStringLiteral("setOffMs"), 0);
+}
--- a/plugins/Lights/HfdLights.h
+++ b/plugins/Lights/HfdLights.h
@@ -21,12 +21,7 @@
 
 #include <QtCore/QObject>
 #include <QDBusInterface>
-#include <QtGui/QColor>
 #include "Lights.h"
-
-#define HFD_SERVICE_NAME "com.lomiri.hfd"
-#define HFD_SERVICE_PATH "/com/lomiri/hfd"
-#define HFD_SERVICE_INTERFACE "com.lomiri.hfd.Leds"
 
 class HfdLights: public Lights
 {
--- a/plugins/Lights/HfdLights.h
+++ b/plugins/Lights/HfdLights.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2021 UBports Foundation.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Florian Leeber <florian@ubports.com>
+ */
+
+#ifndef UNITY_HFDLIGHTS_H
+#define UNITY_HFDLIGHTS_H
+
+#include <QtCore/QObject>
+#include <QDBusInterface>
+#include <QtGui/QColor>
+#include "Lights.h"
+
+#define HFD_SERVICE_NAME "com.lomiri.hfd"
+#define HFD_SERVICE_PATH "/com/lomiri/hfd"
+#define HFD_SERVICE_INTERFACE "com.lomiri.hfd.Leds"
+
+class HfdLights: public Lights
+{
+    Q_OBJECT
+
+public:
+    explicit HfdLights(QObject *parent = 0);
+
+protected:
+    bool init() override;
+    void turnOff() override;
+    void turnOn() override;
+
+private:
+    QDBusInterface* m_hfdInterface;
+};
+
+#endif
--- a/plugins/Lights/LegacyLights.cpp
+++ b/plugins/Lights/LegacyLights.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2014 Canonical, Ltd.
+ * Copyright (C) 2021 UBports Foundation.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/plugins/Lights/LegacyLights.cpp
+++ b/plugins/Lights/LegacyLights.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2014 Canonical, Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Renato Araujo Oliveira Filho <renato.filho@canonical.com>
+ */
+
+#include "LegacyLights.h"
+#include <QtCore/QDebug>
+
+extern "C" {
+#include "android-hardware.h"
+#include <string.h>
+}
+
+LegacyLights::~LegacyLights()
+{
+    if (m_lightDevice) {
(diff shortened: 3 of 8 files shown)


peat-psuwit:
I don't particularly like tack-on override like this. At first, I would asked you to add another class implementing the same interface, and do detection at QML singleton providing time.

Although, there already exists the Hfd implementation in Unity8:
https://github.com/ubports/unity8/commit/e236c3f810c844bbe096c24eb8abb22bf595e646

It completely replaces the old backend with Hfd, using Hfd's QML plugin. Alternatively, one could use Hfd's QML plugin in addition to the current plugin. Although, note that Hfd is suppose to work on older devices too, so we might as well consider doing the transition to Hfd now...

@mariogrip, what do you think? What's the current state of Hfd?

Flohack74:
@peat-psuwit we have LED regression now ongoing since months if not already for a whole year. I agree that this solution is not optimal, but at least it brings back functionality that is missing from the product.
The other PR looks nice, but will it work on old devices? Do we have again something thats on a branch and needs to be separately installed for >= H9 and Pinephone?
The magic about continous integration of an ever-changing feature world is to make small, sometimes silly changes that might need refactor later - which can be expensive, but with the beauty of keeping users enabled and functions functioning.
I find it really a poor approach that hfd was dropped in, stayed unfinished and the consumers around were partly adjusted to it. In the end Halium 9 was regressing from the state in H7.1 and 5.1 in at least vibrator and LEDs API a lot, and I dont think we are serving the users by keeping the fucntionality broken until we find the perfect long-term solution.
Software development in 2021 is no longer a waterfall, but a series of iterations that should ideally never do breaking changes, work a lot with feature flags and parallel implementations, and always strive to prevent long-lived feature branches: Modify, Integrate test, rinse & repeat.

Flohack74:
I also can add that what @mardy said in https://github.com/ubports/hfd-service/issues/22 is true and I dont like to have to call 5 times into DBus for enabling a darn LED, and yet I still would use the current hfd API for LEDs to reach my goal. I could also say we need to refactor whole hfd and make better APIs inside, and until we reach that goal nobody should use it, but thats the wrong approach. The right approach is: Think about the usability of the product, and see it from the point of a user. Users will not care for the technical solution so much but want a result. So, give them a result, and then lets make it better. And dont make breaking changes that stay around for months and years.

peat-psuwit:
If I understand it correctly, the issue affects Android 10 ports and Android 9 ports lacking legacy .so-based Light HAL. Have LED ever worked on those? If it has never it's not actually a regression...

> The other PR looks nice, but will it work on old devices?

I believe it should, given that Hfd-service is included in the image.

And, if you prefer to keep the old code, I'm not opposed to that. What I'm asking is a better way to integrate the new code it. Either:
- Write a new class implementing the same interface, and then do the detection and instantiate the correct class [here](https://github.com/ubports/unity8/blob/7b30fe3512d09b4eb0a8cdca5f94bbcab5c7766a/plugins/Lights/plugin.cpp#L28). You don't even have to do the subclass/virtual method; it'll be handled at QObject level. Or:
- Better yet, find out if we can instantiate the Hfd's QML plugin, or somehow make the QML file dynamically use the correct QML plugin. I'm not sure if it'll make the code more ugly or not though...

Flohack74:
Well it might have never worked for those devices, but as a user I don't care for the low-level peculiarities: If I am a user changing a device from an old Nexus 5 to a newer device then it will feel like a regression. Technically its not a regression of that kind what you think, but its an incomplete implementation of a feature that drags on, and in my opinion that's even worse: Something was started, but then, for whatever reason, energy was lost to complete it. And that happens mostly when one task is too big to be solved in a reasonable time frame, including testing and documentation, if possible. Its about slicing and scoping out until you get an implementable unit of functionality that can be easily integrated.
Well with my limited time this is off the plate for me. Someone else need to pick that up then. I have a solution that works for me, to get me LED notifications back, and I can install this branch if I update my device.

plugins/Lights/Lights.cpp
```diff
@@ -110,10 +126,23 @@ void Lights::setOffMillisec(int offMs)
...
 bool Lights::init()
 {
-    if (m_lightDevice) {
+    if (m_usesHfd || m_lightDevice) {
+        return true;
+    }
+
+    //Try first to detect if hfd should be used
+    QProcess getpropProcess;
+    qWarning() << "Testing for availability of hfd interface";
```
amartinz: Should this really be a warning or rather be something less verbose?

plugins/Lights/Lights.cpp
```diff
@@ -110,10 +126,23 @@ void Lights::setOffMillisec(int offMs)
...
+    getpropProcess.start("/usr/bin/getprop", QStringList() << "ro.build.version.sdk");
+    getpropProcess.waitForFinished();
+    int processResult = getpropProcess.readAllStandardOutput().toInt();
+    qWarning() << "getprop ro.build.version.sdk result:" << processResult;
```
amartinz: Same as above qWarning

plugins/Lights/Lights.cpp
```diff
@@ -110,10 +126,23 @@ void Lights::setOffMillisec(int offMs)
...
+    if (processResult == 0 || processResult >= 27) {
+        m_usesHfd = true;
         return true;
     }
 
+    //else try to initialize a legacy hw module for the LEDs
```
amartinz: `// else`

plugins/Lights/Lights.cpp
```diff
@@ -110,10 +126,23 @@ void Lights::setOffMillisec(int offMs)
 
 bool Lights::init()
 {
-    if (m_lightDevice) {
+    if (m_usesHfd || m_lightDevice) {
+        return true;
+    }
+
+    //Try first to detect if hfd should be used
```
amartinz: `// Try`

amartinz:
And i agree to Florian's point of view.
We should rather unblock users now and be more agile, especially with all the limited resources we have.

End users care if something works and not how technically perfect it is implemented, because most of them are not software architects and bother diving into the code.

From a users point of view all they see "They can not even make a simple light blink, how should i have faith in them being able to manage a whole operating system and trust them with my data when using it as daily driver?"
But that is a whole different discussion and off-topic.

The changes itself look fine and leightweight to me and they are not intrusive.
If there are bigger plans for a certain direction, they can always be pursued but i would prefer to get things in a working state as soon as possible instead of postponing and hoping that we ever get the resources, time and motivation to work on it.

plugins/Lights/Lights.cpp
```diff
@@ -110,10 +126,23 @@ void Lights::setOffMillisec(int offMs)
...
+    QProcess getpropProcess;
+    qWarning() << "Testing for availability of hfd interface";
+    getpropProcess.start("/usr/bin/getprop", QStringList() << "ro.build.version.sdk");
+    getpropProcess.waitForFinished();
+    int processResult = getpropProcess.readAllStandardOutput().toInt();
```
peat-psuwit: We currently has to link against libhybris anyway. Use `property_get()`.

You probably has to link against `libandroid-properties`. See [this file](https://github.com/ubports/unity8/blob/9ec9925db70d3fb34759cf81ce1272e84176a412/src/CMakeLists.txt).

plugins/Lights/Lights.cpp
```diff
@@ -136,6 +165,15 @@ bool Lights::init()
...
 void Lights::turnOn()
 {
+    if (m_usesHfd) {
+        m_hfdInterface->call(QStringLiteral("setState"), 0);
+        m_hfdInterface->call(QStringLiteral("setColor"), (quint32)m_color.rgba());
+        m_hfdInterface->call(QStringLiteral("setOnMs"), m_onMs);
+        m_hfdInterface->call(QStringLiteral("setOffMs"), m_offMs);
+        m_hfdInterface->call(QStringLiteral("setState"), 1);
+        return;
+    }
```
peat-psuwit: 2 branches of equal importance should take a form of:

```C++
if (m_usesHfd) {
  ...
} else {
  ...
}
```

plugins/Lights/Lights.cpp
```diff
@@ -152,6 +190,14 @@ void Lights::turnOn()
 
 void Lights::turnOff()
 {
+    if (m_usesHfd) {
+        m_hfdInterface->call(QStringLiteral("setState"), 0);
+        m_hfdInterface->call(QStringLiteral("setColor"), (quint32)0);
+        m_hfdInterface->call(QStringLiteral("setOnMs"), 0);
+        m_hfdInterface->call(QStringLiteral("setOffMs"), 0);
+        return;
+    }
```
peat-psuwit: Ditto.

plugins/Lights/Lights.h
```diff
@@ -62,10 +63,12 @@ class Lights: public QObject
 
 private:
     light_device_t* m_lightDevice;
+    QDBusInterface* m_hfdInterface;
     QColor m_color;
     State m_state;
     int m_onMs;
     int m_offMs;
+    bool m_usesHfd;
```
peat-psuwit: IIRC, POD type doesn't initialize automatically. Set the default value here, or in initializer list near the constructor, or the constructor itself.

Or, actually, consider eliminating this variable, and initialize `m_hfdInterface` in `init()` instead.

peat-psuwit (changes requested):
Look, I'm not opposed to unblocking the users. But this is a code review, where we should care about readability and maintainability at least a bit. Otherwise, anyone could commit anything citing "this unblocks user".

Okay, I suppose you're thinking that creating a new class is too much work. That's fine, but at least fix the following, so that it feels less tack-on:

plugins/Lights/Lights.cpp
peat-psuwit: Actually, I guess if we try to initialize the legacy HAL first, and then initialize Hfd interface if failed, we might not have to have this property. Unless there exists a device whose legacy HAL is broken...

Flohack74:
Alright I went for a subclassed approach as this requires equal work than all the review comments probably ;)

plugins/Lights/HfdLights.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "HfdLights.h"
+#include <QtCore/QDebug>
+#include <QProcess>
+#include <QDBusArgument>
+#include <QDBusConnection>
+#include <iostream>
+
+extern "C" {
+#include <string.h>
+}
```
peat-psuwit: Nitpick: `str*()` functions seems not used (anymore). This should be removable.

plugins/Lights/HfdLights.cpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Florian Leeber <florian@ubports.com>
+ */
+
+#include "HfdLights.h"
+#include <QtCore/QDebug>
+#include <QProcess>
```
peat-psuwit: Ditto for `QProcess()`.

plugins/Lights/HfdLights.h
```diff
@@ -0,0 +1,47 @@
...
+#define UNITY_HFDLIGHTS_H
+
+#include <QtCore/QObject>
+#include <QDBusInterface>
+#include <QtGui/QColor>
+#include "Lights.h"
+
+#define HFD_SERVICE_NAME "com.lomiri.hfd"
+#define HFD_SERVICE_PATH "/com/lomiri/hfd"
+#define HFD_SERVICE_INTERFACE "com.lomiri.hfd.Leds"
```
peat-psuwit: Nitpick: I guess this can be moved to the .cpp file, as they're not used outside the class.

plugins/Lights/LegacyLights.cpp
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (C) 2014 Canonical, Ltd.
```
peat-psuwit, line 2: Nitpick: I believe we're entitled for a copyright too (in addition).

plugins/Lights/HfdLights.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Author: Florian Leeber <florian@ubports.com>
+ */
+
+#ifndef UNITY_HFDLIGHTS_H
+#define UNITY_HFDLIGHTS_H
+
+#include <QtCore/QObject>
+#include <QDBusInterface>
+#include <QtGui/QColor>
```
peat-psuwit: Nitpick: there's no need to include these in header files.

plugins/Lights/LegacyLights.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (C) 2014 Canonical, Ltd.
```
peat-psuwit, line 2: Ditto for copyright.

plugins/Lights/LegacyLights.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Author: Renato Araujo Oliveira Filho <renato.filho@canonical.com>
+ */
+
+#ifndef UNITY_LEGACYLIGHTS_H
+#define UNITY_LEGACYLIGHTS_H
+
+#include <QtCore/QObject>
+#include <QDBusInterface>
+#include <QtGui/QColor>
```
peat-psuwit: Ditto for includes.

plugins/Lights/Lights.h
```diff
@@ -20,10 +20,9 @@
 #define UNITY_LIGHTS_H
 
 #include <QtCore/QObject>
+#include <QDBusInterface>
```
peat-psuwit: Ditto for includes.

plugins/Lights/plugin.cpp
```diff
@@ -18,14 +18,33 @@
...
 #include <QtQml/qqml.h>
 
 static QObject *lights_provider(QQmlEngine *engine, QJSEngine *scriptEngine)
 {
     Q_UNUSED(engine)
     Q_UNUSED(scriptEngine)
-    return new Lights();
+
+    bool use_hfd = false;
+    char buffer[20];
```
peat-psuwit: Use `PROP_VALUE_MAX` for the array size. In practice it probably won't exceed, but to be safe.

peat-psuwit (commented):
This looks Ok. Most comments are nitpicks. Tested on bq Aquaris 4.5; at least the charging light works, which means at least the legacy code is not broken.

peat-psuwit (approved):
So, all comments resolved. Re-tested on bq 4.5 and the charging LED still works. I've tested this on a Volla Phone. I can't see a charging LED. However, I'm not sure if this device has an LED or not; issuing command directly to HFD via `gdbus` doesn't show anything either.

So, if (and only if) you re-tested this on another Android 9 device with an LED, you may merge this.

Flohack74:
Yeah, the Vollaphone does not have a LED :) I will re-test since I made substantial changes on my SHIFT 6mq

Flohack74:
Aaand it still works. nice. Lets do it!

cibersheep:
Device: Arale
Channel: RC 2021-W45
Result: Leds seem to work as previously (on notification), tapping on home button when screen is on and when phone is off and charging

Flohack74:
We also need a test from a few devices >= Halium9

amartinz:
I can confirm that it started working for a Halium 10 device with this change, where it did not work previously (axolotl - SHIFT6mq).
But you already know that :smile:

jranaraki:
LED works on Nexus 6P on RC updated on Nov 17th. If it helps, the LED on Nexus 6P never stopped working.

PR #417, opened 2021-10-07
Fix ghost window if application creates and destroys surface early in lifecycle

NotKit:
When surface is destroyed by application, SurfaceManager won't destroy it
immediately if it is displayed. liveChanged(false) signal is propagaded instead
all the way up to ApplicationWindow.qml and TopLevelWindowModel. TopLevelWindowModel
may set removeOnceSurfaceDestroyed = true to remove the window in onSurfaceDestroyed
handler. However it won't get fired until ApplicationWindow unreferences surface in
surfaceContainer.

ApplicationWindow has a surfaceInitialized variable hack which is set by a timer after
100ms the surface is live. So if a surface is destroyed faster than that, it never
sets the variable and does not reach "closed" state which unreferences the surface.

Attempt to fix that by introducing one more "surfaceSeenLive" variable that is set
once surface becomes live and helps to understand if it appeared but died too quickly.

### DIFF
--- a/plugins/WindowManager/TopLevelWindowModel.cpp
+++ b/plugins/WindowManager/TopLevelWindowModel.cpp
@@ -317,10 +317,11 @@ void TopLevelWindowModel::onSurfaceDied(unityapi::MirSurfaceInterface *surface)
 
     auto application = m_windowModel[i].application;
 
-    // can't be starting if it already has a surface
-    Q_ASSERT(application->state() != unityapi::ApplicationInfoInterface::Starting);
+    DEBUG_MSG << " application->name()=" << application->name()
+              << " application->state()=" << application->state();
 
-    if (application->state() == unityapi::ApplicationInfoInterface::Running) {
+    if (application->state() == unityapi::ApplicationInfoInterface::Running
+        || application->state() == unityapi::ApplicationInfoInterface::Starting) {
         m_windowModel[i].removeOnceSurfaceDestroyed = true;
     } else {
         // assume it got killed by the out-of-memory daemon.
--- a/qml/Stage/ApplicationWindow.qml
+++ b/qml/Stage/ApplicationWindow.qml
@@ -53,6 +53,7 @@ FocusScope {
         if (surface) {
             surfaceContainer.surface = surface;
             d.liveSurface = surface.live;
+            d.surfaceSeenLive = d.liveSurface;
             d.hadSurface = false;
             surfaceInitTimer.start();
         } else {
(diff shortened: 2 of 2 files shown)


plugins/WindowManager/TopLevelWindowModel.cpp
```diff
@@ -317,11 +317,14 @@ void TopLevelWindowModel::onSurfaceDied(unityapi::MirSurfaceInterface *surface)
...
-    Q_ASSERT(application->state() != unityapi::ApplicationInfoInterface::Starting);
+    DEBUG_MSG << " application->name()=" << application->name()
+              << " application->state()=" << application->state();
 
-    if (application->state() == unityapi::ApplicationInfoInterface::Running) {
+    if (application->state() == unityapi::ApplicationInfoInterface::Running
+        || application->state() == unityapi::ApplicationInfoInterface::Starting) {
         m_windowModel[i].removeOnceSurfaceDestroyed = true;
+        // remove the window now and unreference the surface without going through QML hops
+        deleteAt(i);
```
NotKit: This is not needed if surfaceSeenLive part is applied in QML. However it seemed more sensible to me to remove window right in onSurfaceDied when needed than go through state change in QML and setting surfaceContainer.surface to null;

qml/Stage/ApplicationWindow.qml
```diff
@@ -251,6 +259,7 @@ FocusScope {
                 name: "splashScreen"
                 when:
                      !d.hadSurface && (!root.surface || !d.surfaceInitialized)
+                     && (d.liveSurface ? d.liveSurface : !d.surfaceSeenLive)
```
mardy: Isn't this equivalent to 
```suggestion
                     && (d.liveSurface || !d.surfaceSeenLive)
```

qml/Stage/ApplicationWindow.qml
```diff
@@ -119,6 +121,12 @@ FocusScope {
                 focusedSurface.focus = true;
             }
         }
+
+        onLiveSurfaceChanged: {
+            if (liveSurface) {
+                surfaceSeenLive = true;
```
mardy: nitpick: I would do this on line 74 in the Connections element, to save one signal connection.

mardy (commented):
Just a couple of minor comments.

qml/Stage/ApplicationWindow.qml
NotKit: My mind was pretty confused on correct condition when I tried to debug this.

mardy (approved):
Code-wise this looks good to me, and it seems a low-risk change anyway.

Flohack74:
Please, do not forget to add this to a project board so it does not slip QA and release notes, and also, please create an issue to link the MR to next time.

Flohack74:
Also please add test instructions for QA team.

cibersheep:
(as asked above) How can we test this?

Flohack74:
@NotKit please give some QA instructions, we need to test this for OTA-20.

Flohack74:
From Nikita: __Hi. This was basically a fix for Waydroid leaving empty window after the session had been started, clicking which would crash the compositor__ - so someone with Waydroid should test this.

c4pp4:
There was three windows on OTA-19 when starting Waydroid - window with android, window with Unity8 label and window with Waydroid logo. Now there are only two, the one with Unity8 label is "missing", so I think it's fixed.

PR #421, opened 2022-01-17
Re-introduce interactive drawer blur

fredldotme:
Allow to set a new 'interactiveBlur' based on a new Android property,
called 'ubuntu.unity8.interactive_blur'.

This should be used to indicate the ability to draw interactive blur
in the drawer quickly without causing too many dropped frames.

On device ports which set the necessary getprop property, allow the
use of interactive blur in the drawer. Use static blur otherwise.

### DIFF
--- a/qml/Launcher/Drawer.qml
+++ b/qml/Launcher/Drawer.qml
@@ -34,6 +34,7 @@ FocusScope {
     readonly property real delegateWidth: units.gu(10)
     property url background
     property alias backgroundSourceSize: background.sourceSize
+    property bool staticBlurEnabled : true
     visible: x > -width
     property var fullyOpen: x === 0
     property var fullyClosed: x === -width
@@ -155,18 +156,21 @@ FocusScope {
 
     Rectangle {
         anchors.fill: parent
-        color: "#111111"
-        opacity: 0.99
+        color: "#BF000000"
 
         Wallpaper {
             id: background
             objectName: "drawerBackground"
+            visible: staticBlurEnabled
+            enabled: staticBlurEnabled
             anchors.fill: parent
             source: root.background
         }
 
         FastBlur {
             anchors.fill: background
+            visible: staticBlurEnabled
+            enabled: staticBlurEnabled
             source: background
             radius: 64
             cached: true
@@ -175,6 +179,8 @@ FocusScope {
         // Images with fastblur can't use opacity, so we'll put this on top
         Rectangle {
             anchors.fill: background
+            visible: staticBlurEnabled
+            enabled: staticBlurEnabled
             color: parent.color
             opacity: 0.67
         }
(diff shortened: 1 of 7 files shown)


qml/Launcher/Drawer.qml
```diff
@@ -155,18 +156,21 @@ FocusScope {
 
     Rectangle {
         anchors.fill: parent
-        color: "#111111"
-        opacity: 0.99
+        color: "#BF000000"
```
Flohack74, line 159: you migh want to use lowercaps color codes since thats kinda standard now.

fredldotme: Can change, it just appeared by reverse-applying the Canonical patch on top of the current state.

Flohack74: Yes please ;)

fredldotme: I could also make color and opacity work the way it was before, since those must have been changed for a reason and are part of what is expected by the static blur parts. Ideally I just add the enabled/visible properties to it.

src/ShellApplication.cpp
```diff
@@ -58,6 +58,12 @@ ShellApplication::ShellApplication(int & argc, char ** argv, bool isMirServer)
 
     m_qmlArgs.setMode(parser.mode());
 
+    {
+        char buffer[200];
+        property_get("ubuntu.unity8.interactive_blur", buffer, "false");
```
Flohack74, line 63: Should we define that rather in the ro.xyz space? As its not changeable at runtime

fredldotme: It's not a security-relevant property, and for debugging purposes by the porter I decided to keep it in the same namespace as aethercast and platform-api booster properties are.

qml/Launcher/Drawer.qml
fredldotme: Just checked again:

1) We need the "BF000000" as otherwise it would turn the transparent one black
2) For consistency across the codebase I'd argue we should keep the uppercase version, since it's uppercase everywhere

Flohack74 (approved):

c4pp4:
RC OTA22: 2022-W04 (devices: Mi A2 and Xperia X)
Not sure If I'm testing it right or I have the right devices. I've done 'setprop ubuntu.unity8.interactive_blur true' and 'restart unity8'. After that when the app drawer is appearing, animation is very choppy.

fredldotme:
> RC OTA22: 2022-W04 (devices: Mi A2 and Xperia X) Not sure If I'm testing it right or I have the right devices. I've done 'setprop ubuntu.unity8.interactive_blur true' and 'restart unity8'. After that when the app drawer is appearing, animation is very choppy.

Yes, this is exactly the reason the porter has to opt-in on the feature instead of the user.
If the porter determines it's not fluid enough then better not enable it. Works fine on Pixel 3a though.

Flohack74:
We might be able to speed it up with better QML later, but for now its truly device-specific ;)

PR #427, opened 2022-02-21
prevent unwanted prompts loading

### LINKED ISSUE #428: Unnecessary prompts loading
Today, if a user have a pinCode prompt, the code will load ButtonPrompt.qml, TextPrompt.qml and finally PinPrompt.qml.
https://github.com/ubports/unity8/blob/xenial/qml/Greeter/GreeterPrompt.qml#L110
Making unnecessary component loading and may lead to focus issue.

lduboeuf:
Today, if a user have a pinCode prompt, the code will load ButtonPrompt.qml, TextPrompt.qml and finally PinPrompt.qml.
Making unnecessary component loading and may lead to focus issue.

the idea is to make sure all properties are set during GreeterPrompt loading to avoid states conditions being true sequentialy

fixes #428

### DIFF
--- a/qml/Greeter/GreeterPrompt.qml
+++ b/qml/Greeter/GreeterPrompt.qml
@@ -25,7 +25,7 @@ FocusScope {
     focus: true
 
     property bool isPrompt
-    property bool isAlphanumeric
+    property bool isPinPrompt
     property string text
     property bool isSecret
     property bool interactive: true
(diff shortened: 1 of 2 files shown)


lduboeuf:
still need to have better state cobdition in GreeterRompt. Some of them can be true at the same time.

Flohack74 (approved):

PR #430, opened 2022-03-18
Fix copyright test

gber:
ButtonPrompt.qml, PinPrompt.qml, TextPrompt.qml were introduced in 4eca819 and
are derived from GreeterPrompt.qml.
clickable.json should be ignored since it is just metadata.

### DIFF
--- a/tests/copyright/check_copyright.sh
+++ b/tests/copyright/check_copyright.sh
@@ -34,9 +34,9 @@ usage()
 # image files
 ignore_pat="\\.sci$|\\.svg$|\\.png$|\\.jpg$"
 # git/bzr files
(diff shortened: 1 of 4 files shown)


tests/copyright/check_copyright.sh
```diff
@@ -65,6 +67,8 @@ ignore_pat="$ignore_pat|\\.qrc$"
 ignore_pat="$ignore_pat|/doc/devices\\.conf$"
 # Jenkins configuration
 ignore_pat="$ignore_pat|/Jenkinsfile"
+# upstart jobs
+ignore_pat="$ignore_pat|/plugins/.*\\.conf$"
```
peat-psuwit: I guess we want to narrow this down to just the Wizard plugin. `.conf` is a somewhat generic extension.

```suggestion
ignore_pat="$ignore_pat|/plugins/Wizard/.*\\.conf$"
```

gber: OK.

PR #432, opened 2022-03-21
qml: Fix Drawer hiding away under the user

fredldotme:
- preventHiding when launcher is in various launcher states

### DIFF
--- a/qml/Launcher/Launcher.qml
+++ b/qml/Launcher/Launcher.qml
@@ -427,7 +427,7 @@ FocusScope {
         Connections {
             target: panel.dismissTimer
             onTriggered: {
-                if (root.autohideEnabled && !root.lockedVisible) {
+                if (root.state !== "drawer" && root.autohideEnabled && !root.lockedVisible) {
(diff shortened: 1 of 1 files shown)


kugiigi:
Not sure if this could be a help but here is my 3am fix :smile: 
I had a different approach. Basically, I found out that the property `autohideEnabled` isn't reset so the launcher/drawer just closes automatically every time after it was opened via pushing the mouse to the left edge at least once.

I wonder if we should also have temporary visible state for the drawer so it will also close once the mouse is out of its area.

Flohack74 (approved):

luksus42:
Not sure, if its connected with this change, but the app drawer is not opening anymore when touching the ubuntu icon in staged mode.

Flohack74:
@fredldotme can you check that last comment?

luksus42:
FYI: I did an update this morning (dev), only 1.2MB update size, and now the Button is working again. So maybe it is already fixed or it was just some device specific issue on my FP3.

PR #438, opened 2022-05-18
increase fontSize for notifications

### LINKED ISSUE #197: Feature request- Display of the contact on call too small
When I am called, the visibility of the phone number or contact name is too small and too weak. I can't see it without my glasses.
Is there a possibility to make this display larger when a call comes in?
The screen is big enough.
        Thank you!
msg-1165295608-82641

lduboeuf:
Some users report that notifications can sometimes be too small to read.
According to my tests, we can't increase a lot without breaking the layout,
so did only one size above.

fixes: https://github.com/ubports/dialer-app/issues/197

screenshot20220517_224228541
screenshot20220517_161713626

Before-> After:
notif-size

### DIFF
--- a/qml/Notifications/Notification.qml
+++ b/qml/Notifications/Notification.qml
@@ -301,7 +301,7 @@ StyledItem {
                             right: parent.right
                         }
                         visible: type !== Notification.Confirmation
-                        fontSize: "medium"
+                        fontSize: "large"
                         font.weight: Font.Light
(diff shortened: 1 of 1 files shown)


lduboeuf:
Well not sure if it is relevant to increase font size only for notification. Shouldn't be more a system wide setting ?

lduboeuf:
should be a system setting

PR #442, opened 2022-06-29
Notch support

ilyakooo0:
This is a rebased version of #403

I don't really understand what is going on, I just rebased the PR (:

### DIFF
--- a/data/com.canonical.Unity8.gschema.xml
+++ b/data/com.canonical.Unity8.gschema.xml
@@ -63,6 +63,11 @@
       <summary>Whether the OSK switch should be visible</summary>
       <description>Toggle the visibility of the OSK switch</description>
     </key>
+    <key type="b" name="disable-top-margin">
+      <default>false</default>
+      <summary>Disable top margin for the notch</summary>
+      <description>This will restore the previous behavior on devices with a notch. If this option is enabled some indicators might not be visible.</description>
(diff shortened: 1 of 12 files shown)


Flohack74:
Nice, we need to clean it a bit, but I definitely want this for 2 distinct phone models now ;)

JamiKettunen:
@Flohack74 Is this "confirmed" to be the way we want to go forward with notch integration on Unity 8 / Lomiri shell? I wouldn't mind making a parallel MR for `focal` on https://gitlab.com/ubports/development/core/lomiri myself ^^

JamiKettunen:
~~With 20.04 (and future) in mind we should refactor `plugins/Utils/deviceconfigparser.*` changes and get a `libdeviceinfo` replacement as that is also available on 16.04 already while [the deviceconfigparser is gone in `focal`](https://gitlab.com/ubports/development/core/lomiri/-/commit/088ce4557b918918d23774b9c50b7c85702ce03b)~~
Ignore me, while `libdeviceinfo` is also on 16.04 we probably don't want to start integrating it into the shell if it already wasn't before, so this might just have to be done differently on 20.04 and it doesn't seem there's backwards compatability for the old deviceconfigparser way so with migration to `focal` existing configuration from `xenial` will have to be migrated to `deviceinfo`.

PR #443, opened 2022-07-07
Fingerprint unlock: backoff time between retries

Flohack74:
Currently Lomiri would try to read a fingerprint, and upon failure immediately retry, giving the user no chance to lift the finger. Also 3 retries is quite too low.
Implement a timer-based backoff that allows the user to correct finger position more easily, and also increase the max failed retries to 5.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -12,7 +12,7 @@ export QT_SELECT=qt5
 DEB_HOST_ARCH ?= $(shell dpkg-architecture -qDEB_HOST_ARCH)
 
 # Skip tests on the archs they are known to be flaky  with current configuration
-testskip_architectures := arm64 powerpc ppc64el s390x
+testskip_architectures := powerpc ppc64el s390x
 
 %:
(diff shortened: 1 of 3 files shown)


Flohack74:
Note that this PR introduces a change in behaviour (hence the changes to unit tests): When the fingerprint unlocking method is locked after 5 consecutive failures we no longer automatically emulate the swipe to get to the PIN or password input. I think this is very confusing.
In a second PR we will move up the hint text a bit from the very bottom, and also show a symbol if fingerprint scanner can be used or not.

j2g2rp:
I've been testing this for last day on Volla x. Looks like Fingerprint reader works fine now.

amartinz (approved):

amartinz:
> In a second PR we will move up the hint text a bit from the very bottom, and also show a symbol if fingerprint scanner can be used or not.

Is that plan documented somewhere trackable?

Flohack74:
@amartinz yes currently here: https://github.com/ubports/unity8/issues/447

PR #444, opened 2022-07-13
Use autoTransform property for rotated wallpaper images

Flohack74:
JPEGs from newer devices are often rotated due to the physical mounting on the PCB, and that needs to be considered when taking a photo from the device as wallpaper.


Flohack74:
Contributes to https://github.com/ubports/ubuntu-touch/issues/2009

maciek134 (approved):

amartinz (approved):

PR #446, opened 2022-07-13
qml: Improve Interactive Blur performance

fredldotme:
- Remove most of the visually redundant shaders for color correction.
- Update the blur in a deferred manner, leaving room for the shell to
  render other QML Items.
- Hide the Stage (!) when fully occluded by the Drawer to avoid overdraw.


Flohack74 (approved):

ilyakooo0:
Can confirm that the app drawer background blur is noticeably faster on Pixel 3a.

PR #454, opened 2022-10-19
qml: Even faster blurred Drawer

fredldotme:
The BackgroundBlur component used to be busy on redrawing the blur source whenever the source rectangle changed. This turned out to be the main reason for weaker devices to have problems drawing with the blur shader enabled.

Instead of changing the source rect based on the Drawer's drag distance, do the following:

- Make the BackgroundBlur stay in place attached to the left-hand side.
- Use a parent item with clipping enabled to cause the blur effect to not show up outside of the Drawer's boundaries.
- Change the blur rectangle to only reference the area of interest.

Fixes blurred Drawer performance on more devices
as tested on the JingPad, as well as the weird
jelly effect during drag which was introduced
in the last performance changeset.

### DIFF
--- a/qml/Launcher/BackgroundBlur.qml
+++ b/qml/Launcher/BackgroundBlur.qml
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2016 Canonical, Ltd.
+ * Copyright (C) 2022 UBports Foundation
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/qml/Launcher/BackgroundBlur.qml
+++ b/qml/Launcher/BackgroundBlur.qml
@@ -35,10 +35,7 @@ Item {
 
     FastBlur {
         id: fastBlur
-        x: blurRect.x
-        y: blurRect.y
-        width: blurRect.width
-        height: blurRect.height
+        anchors.fill: parent
         source: shaderEffectSource
         radius: units.gu(3)
         cached: false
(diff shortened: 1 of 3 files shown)


qml/Launcher/Launcher.qml
```diff
@@ -366,17 +366,26 @@ FocusScope {
...
-        occluding: (drawer.width == root.width) && drawer.fullyOpen
+    Item {
+        clip: true
+        x: 0
+        y: drawer.y
+        width: drawer.width + drawer.x
+        height: drawer.height
+        BackgroundBlur {
+            id: backgroundBlur
+            x: 0
```
maciek134, line 377: x/y should default to zero already, maybe this is not needed?

maciek134 (approved):
The x/y stuff would be nice to check, but is basically just style, so doesn't matter that much :)

maciek134:
Verified on the Nexus 5 and 4, no noticeable lag, I guess the only slower device could be the BQ4.5? Similar CPU with more L1 cache than N4, but much slower GPU.

qml/Launcher/Launcher.qml
fredldotme: I believe I did it this way to make it rather obvious and explicit how things work rather than expecting eventual readers to understand it immediately.

maciek134:
CI seems to be failing with `Target 'tests/CMakeFiles/xvfballtests.dir/all' not remade because of errors.`, but no details and no other failed tests.

qml/Launcher/BackgroundBlur.qml
```diff
@@ -35,10 +35,7 @@ Item {
```
Flohack74, line 36: Plz add additional copyright at beginning of file

fredldotme: Added

Issue #351, opened 2020-12-30
Closing app drawer with super key/ubuntu button makes the current app unresponsive


kugiigi:
## Steps to Reproduce
1. Open at least 2 apps
2. Open the app drawer however you want (super key, touch, mouse)
3. Close the app drawer by either pressing the super key on a physical keyboard or clicking on the ubuntu button.

## Expected Results
App drawer closes and the current app works as is

## Actual Results
App drawer closes but the current app becomes unresponsive until you reopen and close the drawer in different way or swicth to other app.

cibersheep:
- I can confirm this behaviour with the bottom physical button on Arale
- Tapping on the Ubuntu logo to dismiss the Drawer:
  - hides the Drawer showing the launcher
  - first tap hides the launcher (foreground app is not responding)
  - second tap on the foreground app (app is responding)

PR #455, opened 2022-10-20
qml: Unfocus launcher when dismissing Drawer

fredldotme:
Until now, when pressing the Super key on a keyboard, the Launcher's focus state was still set to true, causing windows not to receive input events.

Unfocus the Launcher in order to cause the Stage to re-evaluate the new focus state, leading to windows receiving events again.


Flohack74 (approved):

kugiigi:
Is this a fix for #351 ? So maybe that could be closed as well as my old PR :sweat_smile:

Flohack74:
No clue @kugiigi can you test it?

PR #457, opened 2022-12-03
YUMI-7: Fix race condition in XdgWatcher

maciek134:
Happens when a desktop file is deleted and re-created before XdgWatcher::onFileChanged is called - instead of an appRemoved + appAdded, it emitted appInfoChanged + appAdded, resulting in duplicate apps (Waydroid)


JamiKettunen:
Definitely needs a `focal` counterpart against https://gitlab.com/ubports/development/core/lomiri assuming this helps (testing on `xenial` via `sudo ubports-qa install PR_unity8_457`) :)

Flohack74 (approved):

PR #458, opened 2023-01-17
[DO-NOT-MERGE] : test zero emblem

lduboeuf:


lduboeuf:
not satisfying solution yet as app is pinned even if emblem is not visible and app is not running

